PR #645, opened 2026-04-29
Match pending sends by content on echo

meawoppl:
## Summary
The optimistic send drain was blindly removing the first pending message when any user echo arrived. If a message got lost (network drop, server error), the next echo would incorrectly consume an unrelated pending entry, hiding a real message.

Now the echo's `content` field is compared against each pending send's `content`. Only a matching entry is removed. If no match is found (e.g., a message from another user in a shared session), the pending queue is left unchanged.

## Test plan
- [ ] Verify normal send/echo still works (pending message appears, disappears on echo)
- [ ] Verify messages from other users in shared sessions don't consume pending entries
- [ ] Verify multiple rapid sends drain in correct order

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.4.20
+
+- Match pending sends by content when draining on server echo
+
 ## 2.4.19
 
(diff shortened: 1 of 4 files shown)


meawoppl:
Superseded by #659 which combines content-matching with assistant/result clearing for slash command support

Issue #703, opened 2026-05-15
Codex session stuck after upstream frame fails typed decode (`missing field `callId``)


meawoppl:
## Symptom

A live codex session in the launcher became unresponsive: every subsequent user prompt was silently dropped and the proxy logged `Received input while Codex turn is active` even though the underlying `codex app-server` had already finished the turn and was idle (`futex_do_wait`).

Observed on the running launcher (`agent-portal --no-update`, workspace `v?` current `main`), session id `50f6dad6-42bd-4e7c-93ff-f7e998a60fd7`, working dir `~/repos/fitsio-pure`, `codex-codes` v0.101.1.

## Root cause

`claude-session-lib/src/session.rs:724-811` drives a `turn_active` state machine. While `turn_active = true`, the loop pumps `client.next_message()` and only resets the flag when `handle_codex_server_message` returns `turn_ended = true` — which happens **only** in the `turn/completed` arm (lines 829-843).

The `codex-codes` client encountered a frame it couldn't deserialize into its typed `ServerMessage` enum (`missing field \`callId\``), emitted a warning, and **skipped** the frame. The skipped frame was the envelope carrying `turn/completed` (or a sibling notification immediately preceding it), so:

- The codex app-server thinks the turn ended cleanly.
- agent-portal still has `turn_active = true`.
- All subsequent `IoCommand::Input` are dropped at `session.rs:762-764`.
- The session is wedged until the launcher is restarted.

## Evidence (journald, `agent-portal` unit)

```
19:11:31  ws recv: OutputAck seq=35
19:11:33  WARN claude_session_lib::session:
          Codex frame failed typed decode (skipping): missing field \`callId\`
19:11:33  ws recv: OutputAck seq=36
19:11:33  Codex notification: account/rateLimits/updated
19:12:21  ws recv: OutputAck seq=37
19:14:04  SequencedInput seq=3 content=\"is it working?\"
19:14:04  WARN Received input while Codex turn is active     ← stuck
19:16:59  SequencedInput seq=4 content=\"workin?\"
19:16:59  WARN Received input while Codex turn is active     ← still stuck
```

(The `failed typed decode` warning is emitted from inside the `codex-codes` async client, not from agent-portal — confirmed via journald target `claude_session_lib::session` re-emitting stderr from the dependency.)

## Recovery (today)

[...]

PR #708, opened 2026-05-15
Revert #707 CodexFrameCaptureLayer; wait on SDK issue #128

meawoppl:
## Summary

Pulls out the tracing-subscriber workaround from #707 (2.5.30). It didn't actually capture frames in production — the per-callsite `Interest::always()` / per-layer EnvFilter rewiring interacted with tracing-subscriber 0.3 in a way the unit test didn't exercise, so the ring buffer stayed empty and the portal message always fell through to the "Raw frame was not captured" branch.

Rather than chase that down, [rust-code-agent-sdks#128](https://github.com/meawoppl/rust-code-agent-sdks/issues/128) is being implemented: `codex-codes` will expose a `ParseError` carrying `raw_line` / `raw_json` / `error_message`, mirroring claude-codes' existing pattern. Once that ships we'll match on `codex_codes::Error::Deserialization(parse_err)` and render `raw_json` straight to the portal message — the way the claude path has handled unparsable frames forever.

## What's removed

- `claude-session-lib/src/codex_frame_capture.rs` (the layer module).
- The layer install + per-layer EnvFilter rewiring in `launcher/src/main.rs` and `proxy/src/main.rs` (back to the pre-#707 registry-level EnvFilter).
- `tracing-subscriber` from `claude-session-lib`'s deps (only #707 needed it).
- The raw-frame portal message + fallback note in `session.rs`.

## What stays

The #703 / #704 / 2.5.26 recovery is unchanged — on `codex_codes::Error::Json` we still emit the short `turn.failed` portal message and call `turn/interrupt` so codex unblocks. The user keeps an unwedged session and clear visibility; they just don't get the raw frame attached until the SDK fix lands.

## Test plan

- [x] `cargo check --workspace`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] `cargo test -p claude-session-lib`
- [x] Manual: dead-code grep — no stale `codex_frame_capture` / `CodexFrameCaptureLayer` / `take_most_recent` references remain.


meawoppl:
Superseded by #720, which absorbs codex-codes 0.129.3 (SDK #134's structured ParseError landed there). The 0.129.3 bump deletes the CodexFrameCaptureLayer and switches session.rs to match on `Error::Deserialization(parse_err)` directly, which is what this PR was waiting for.

PR #789, opened 2026-05-18
Codex item-lifecycle dedup: collapse running/completed cards (closes #776)

### LINKED ISSUE #776: Codex bash command: condense running → completed into a single message instead of two duplicates
Reported by user 2026-05-18. When codex runs a bash command we currently render the command twice in the transcript:

1. A "running" message when the command starts.
2. A "completed" message with the output when it finishes.

The two cards look near-identical (same command text), which is visually noisy ("fugly"). Should be collapsed into a single mutable card that updates in-place from "running" → "completed" (or with a spinner that flips to a checkmark + duration + output once done).

**Where**: codex's `item/commandExecution/*` / `execCommand` event sequence in `frontend/src/components/codex_renderer.rs`. Likely we're treating `ItemStarted` and `ItemCompleted` as independent transcript entries when we should be folding the completion into the started item's slot (keyed by `item_id`).

**Look at**: the streaming-delta renderers added in #737 already do in-place updates for `turn/diff/updated` / `turn/plan/updated` — same pattern should apply here. The exec-command event shape may need its own `CodexEvent` variant if the typed dispatch isn't already routing to a single updatable card.

**Acceptance**: a single transcript card per bash command that goes from spinner+"running" to checkmark/X+"completed in Xs" with the output, no duplicate.

meawoppl:
## Summary

A single codex bash command lands on the wire as `item.started` (status: running) immediately followed by `item.completed` (status: completed, with output + exit code) — and any number of `item.updated` events in between when output streams. Before this fix, each landed as a separate transcript card, so a `cargo test` invocation showed up as two near-identical "Codex / commandExecution" cards (running + completed) side-by-side. A long-streaming MCP call could show many.

## Fix

`render_codex_group` now runs a `codex_renderer::dedupe_lifecycle` pass first:
- Parses each raw JSON into the typed `CodexEvent` (per #767).
- Extracts a `(item_type_tag, item_id)` key off the `ItemStarted | ItemUpdated | ItemCompleted` arm's `CodexItem`.
- Collapses runs of the same key to only the latest occurrence (the `ItemCompleted` when the item finished; the latest in-flight `ItemUpdated` when it's mid-stream; the lone `ItemStarted` for a hung command).

The surviving message sits at the position of its latest occurrence so the transcript reflects when the item finished, not when it started.

**Passthrough cases** (deliberately not collapsed):
- Non-lifecycle events: TurnCompleted, the streaming-delta variants from #737, etc.
- Lifecycle events on `CodexItem::Unknown` (no id available).
- Malformed-JSON entries — we never silently drop a message.

**Dedup key includes the type tag** — a `command_execution` with `id="a"` and a `file_change` with `id="a"` correctly stay separate.

**Live-stream UX is also better**: the in-progress card is visible until completion arrives, then naturally replaced on the next re-render. Old session history reloaded from the DB is cleaned up retroactively for free.

## Test plan

- [x] Six new unit tests in `codex_renderer::tests` cover the motivating bash regression, multi-update streaming runs, independent items not collapsing, non-lifecycle passthrough, items-without-id passthrough, and malformed-JSON passthrough.
- [x] `cargo check --workspace --exclude backend` clean
- [x] `cargo fmt --all` clean
- [x] `cargo clippy --workspace --exclude backend -- -D warnings` clean
- [x] `cargo test -p frontend --lib codex_renderer` — all 38 codex_renderer tests pass


meawoppl:
Superseded: the #776 item-lifecycle dedup (collapse running→completed codex cards) already shipped on main via `grouping::visible_group_indices` (suppresses earlier events sharing an `item_id` with a later event in the same group). This branch also conflicts with the codex-renderer refactors (#803–#809) since. Closing as done-by-other-means; reopen if there's behavior here not covered by `visible_group_indices`.

Issue #788, opened 2026-05-18
History APIs load all messages before frontend trimming


meawoppl:
## Finding (static review)

REST \`list_messages\` loads the full session history (in \`messages.rs\`), and the WebSocket replay path does the same when no \`replay_after\` is supplied (in \`web_client_socket.rs\`). The frontend then truncates locally. So for long sessions, the entire history crosses the wire and only the tail gets shown.

## Affected files

- \`backend/src/handlers/messages.rs\` (\`list_messages\`)
- \`backend/src/handlers/web_client_socket.rs\` (no-\`replay_after\` branch)

## Suggested fix

Push pagination / limits into SQL:
- REST: add \`limit\` + \`before\`/\`after\` query params, default to the frontend's render-window size.
- WS: when \`replay_after\` is absent, default to last N messages instead of the whole history.

Should also help with the load-time TTFB of long sessions.

## Severity

Performance — wasted bandwidth + memory on every reconnect / initial load of a long session.

## Source

Filed from a static-review pass.

PR #800, opened 2026-05-18
Reconnect replay uses server-assigned timestamps (closes #784)

### LINKED ISSUE #784: Reconnect replay can miss messages due to client-side timestamps
## Finding (static review)

The frontend sets \`last_message_timestamp\` to browser \"now\" (in \`component.rs\`), while the backend filters DB messages with \`created_at.gt(after)\` (in \`web_client_socket.rs\`). Clock skew between client and server, or messages already in flight, can create replay gaps that the user never sees.

## Affected files

- \`frontend/src/components/.../component.rs\` (sets \`last_message_timestamp\` from \`Date.now()\` or similar)
- \`backend/src/handlers/web_client_socket.rs\` (\`created_at.gt(after)\` filter)

## Suggested fix

Send server-assigned \`created_at\` / sequence IDs back with each live message. Have the frontend remember the last server-assigned ID it actually rendered, and replay strictly from that on reconnect. Don't trust client clocks for replay watermarks.

## Severity

Data loss on reconnect — silent gaps that look like normal stream behavior.

## Source

Filed from a static-review pass.

meawoppl:
## Summary

- Silent data-loss fix on the WebSocket reconnect path: the frontend used to set `last_message_timestamp` from `js_sys::Date::new_0()` (browser "now") and replay from that on reconnect; backend filters `messages.created_at.gt(after)`, so clock skew or messages in flight at the watermark moment silently dropped messages the user never saw.
- Wire-shape change (minimal, `#[serde(default)]` everywhere for backward compat): `ServerToClient::ClaudeOutput` gains `created_at: Option<String>` sourced from the persisted DB row, and `ServerToClient::HistoryBatch` gains a matching `last_created_at: Option<String>` summary field.
- Backend reorders insert-before-broadcast and uses `get_result::<Message>` to recover the row's server-assigned `created_at`; `replay_history` also injects `_created_at` per batch entry (matching the REST path). Frontend `WsEvent::Output` / `HistoryBatch` carry the timestamp through, component's two `Date.new_0()` watermark sites are replaced with conditional updates from the server-supplied value.

## Test plan

- [x] `cargo check --workspace --exclude backend`
- [x] `cargo build -p backend`
- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` (clean)
- [x] `cargo test -p shared` (42 tests pass; wire-compat for pre-#784 frames)
- [x] `cargo test -p frontend --lib` (121 tests pass; 5 new for WS→WsEvent translation including the server-timestamp round-trip + the error-envelope non-DB path)
- [x] `cargo test -p backend` w/ `DATABASE_URL` set (70 tests pass; 2 new DB-backed `replay_history_db_tests` pin the strict-`>` contract + high-water advance — auto-skip when DB unset, same pattern as `session_access::db_tests`)

Manual verify pending: reconnect on the dev stack with a deliberately-skewed browser clock to confirm a previously-dropped in-flight message now arrives in the replay batch.


meawoppl:
Closing — files conflicted with #788 (history pagination) which merged first. Will be re-opened against the current main.

PR #1108, opened 2026-06-24
Track Codex subagent token metrics

meawoppl:
## Summary
- attribute Codex child-thread token deltas into TurnMetrics.subagent_tokens while keeping main-thread token usage separate
- expose thinking/subagent token sums in the metrics aggregate API
- surface auxiliary token usage in the turn footer, dashboard metric pill, and Settings > Performance chart

## Notes
- Uses typed Codex app-server events: ThreadStarted identifies child threads via parent_thread_id, and ThreadTokenUsageUpdated.last supplies per-update token deltas.
- SDK tracking issue for Claude rollup/subagent fields: https://github.com/meawoppl/rust-code-agent-sdks/issues/168

## Verification
- cargo fmt --check
- cargo test -p codex-session-lib -p shared
- cargo test -p frontend
- cargo build -p shared --target wasm32-unknown-unknown
- env -u NO_COLOR trunk build
- cargo test -p backend
- cargo clippy --workspace --all-targets -- -D warnings

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agent-portal"
-version = "2.12.20"
+version = "2.12.21"
 dependencies = [
  "anyhow",
  "chrono",
@@ -219,7 +219,7 @@ dependencies = [
 
 [[package]]
 name = "backend"
-version = "2.12.20"
+version = "2.12.21"
 dependencies = [
  "anyhow",
  "axum",
@@ -487,7 +487,7 @@ dependencies = [
 
 [[package]]
 name = "claude-portal"
-version = "2.12.20"
+version = "2.12.21"
 dependencies = [
  "anyhow",
  "base64",
@@ -518,7 +518,7 @@ dependencies = [
 
 [[package]]
 name = "claude-session-lib"
-version = "2.12.20"
+version = "2.12.21"
 dependencies = [
  "anyhow",
  "base64",
@@ -556,7 +556,7 @@ dependencies = [
 
 [[package]]
 name = "codex-session-lib"
-version = "2.12.20"
+version = "2.12.21"
 dependencies = [
  "chrono",
  "claude-codes",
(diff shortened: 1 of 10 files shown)


meawoppl:
Reviewed — overall this is clean, idiomatic, and well-tested. The key design call is right: filtering main-thread usage to `thread_id == thread_id && turn matches` and routing child-thread usage to a separate tracker means the main `input/output/cache/thinking` counts and `subagent_tokens` don't double-count each other. Backend `::bigint` casts on the new SUMs match the `#[diesel(sql_type = BigInt)]` AggregateRow fields, `MetricBucket` additions are `#[serde(default)]` (back-compat), and the footer/pill/Performance-chart surfaces all have tests. Nice.

**One substantive question before I merge — subagent accumulation semantics.** The main path *latches* `last` (last-wins): `current_turn_usage = Some(p.token_usage.last.clone())` on each main-thread update. The subagent path *sums* it: `current_turn_tokens += token_breakdown_total(&p.token_usage.last)` across every child-thread update.

That asymmetry is only correct if `ThreadTokenUsageUpdated.last` is emitted **at most once per (thread, turn)** — i.e. it's effectively a per-turn delta. If codex *streams* `last` multiple times within a single child turn (which is exactly why the main path latches rather than assigns), then summing a child thread's updates over-counts: you'd add the intermediate growing values, not just the turn total.

If that's a real risk, the robust shape is to accumulate per child thread keyed by `(thread_id, turn_id)` with last-wins, then sum across threads at `take` — or use `total` (cumulative) and take end-of-parent-turn minus start-of-parent-turn. Both are immune to update cadence.

Can you confirm `.last` cadence for child threads? If it's once-per-turn, summing is correct and I'll merge as-is. If it streams, a quick switch to last-wins-per-(thread,turn) before merge would keep the headline metric accurate. Everything else is 👍.

meawoppl:
Verified the fix at bf9fdaf — `observe_token_usage` now keys by `(thread_id, turn_id)` with last-wins (`.insert` overwrites), and `take` sums across distinct child-turn entries. That's cadence-immune: repeated `.last` updates within one child turn overwrite instead of accumulate, exactly matching the main path's latest-usage semantics, while genuinely distinct child turns/threads still each contribute. Regression test covers the repeated-update case. My one concern is resolved — approving. Will merge once CI (incl. container) is green. 🚢

PR #1109, opened 2026-06-25
Extract SessionView buffer state helpers

meawoppl:
## Summary
- Extract SessionView message retention and turn-metrics buffer rules into `session_view/state.rs`
- Reuse the helpers from history replay, REST hydration, live websocket output, and live metrics insertion
- Add focused tests for newest-item retention, live append trimming, sorted metric insertion, id deduplication, and hydration sorting
- Bump workspace version to 2.12.22

## Verification
- `cargo fmt --check`
- `cargo test -p frontend`
- `cargo clippy -p frontend --all-targets -- -D warnings`

Frontend-only; intentionally avoids `shared/`, `session-lib/`, `claude-session-lib/`, and backend while Claude works on the adapter fold.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agent-portal"
-version = "2.12.21"
+version = "2.12.22"
 dependencies = [
  "anyhow",
  "chrono",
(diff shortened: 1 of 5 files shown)


meawoppl:
Reviewed — clean extraction. The helpers are pure and well-tested without mounting Yew, the invariants are documented (newest-tail retention, id-dedup keeping `None`-id rows distinct, `started_at ASC` via binary-search insert), and `insert_turn_metrics_sorted`'s binary search stays correct because every insertion path keeps the buffer sorted (live insert + hydration via `sort_turn_metrics_by_start`). Frontend-only, CI green, 2.12.22 as agreed. LGTM 🚢

PR #1110, opened 2026-06-25
Stabilize session pill ordering + focus for same-repo agents (#1094)

### LINKED ISSUE #1094: Stabilize session pill ordering for multiple agents in the same repo
## Problem

When multiple agent sessions are running against the same repository, especially via the intended messaging/worktree workflow, the dashboard session pills can appear to reorder unpredictably. This shows up as weird focus bounces: the highlighted pill and focused session view can jump even though the user did not intentionally switch sessions.

This undermines multi-agent same-repo support because the user needs the rail order and focused session to remain stable while several agents are active in nearby worktrees.

## Current Triage

Likely contributing factors:

- `frontend/src/pages/dashboard/page.rs` derives `active_sessions` by sorting only on extracted folder name, then hostname. Multiple agents in the same repo/worktree family on the same host can compare equal. Rust sort order is not a product guarantee for equal keys, and new `/api/sessions` poll responses can arrive in a different source order.
- Focus is stored as `focused_index: usize` in `frontend/src/pages/dashboard/page.rs`, and `SessionRail` receives/returns indices. If the sorted vector order changes, the same index can now refer to a different session.
- `frontend/src/pages/dashboard/session_rail.rs` uses `data-index` and `focused_index` to scroll the focused pill into view. Any index drift can trigger scroll/focus churn.
- `backend/src/handlers/sessions.rs::list_sessions` orders by `sessions::last_activity.desc()`, which is activity-sensitive and can also reshuffle active agents between polling intervals before the frontend applies its partial sort.
- Local metadata updates such as branch/PR changes mutate `sessions` and then re-run the same index-based focus logic, so ties in the comparator can surface as focus instability.

## What We Need To Support

- Multiple simultaneous sessions with identical or near-identical `working_directory` folder names.
- Multiple sessions on the same `hostname`.
- Multiple worktrees for the same Git repository, where `repo_url` may match but `working_directory`, `git_branch`, agent type, and session id differ.
- Polling, branch metadata updates, activity updates, and connection state changes must not reorder equal-looking pills in a way that changes focus.
- Keyboard navigation, nav-mode number badges, hidden/paused sections, and `scroll_into_view` should operate on a stable display model.

## Proposed Fix Shape

1. Track focus by `session_id` as the source of truth, not by array index. Derive the current index from the displayed session list only at render/navigation boundaries.
2. Introduce a total, deterministic session display comparator. Candidate key order:
   - normalized repo/display folder
   - normalized `repo_url` or full `working_directory`
   - `git_branch`
   - `agent_type`
   - `hostname`
   - `created_at` or `session_name`
   - `id` as final tie-breaker
3. Consider preserving user-visible rail order across polling by merging refreshed session data into an existing order map, then inserting truly new sessions at a deterministic position.
4. Update `SessionRail` and keyboard navigation APIs to pass/select `Uuid` where practical, or keep indices strictly as derived display indices with an invariant that focus is re-resolved by id after every session-list update.
5. Add unit tests for the comparator and focus retention behavior:
   - same folder + same host + different session IDs has stable ordering
   - same repo with two worktrees and different branches has stable ordering
   - focused session remains focused after `/api/sessions` returns the same sessions in a different order
   - hidden/paused split does not change the identity of the focused session

## Acceptance Criteria

- Running two or more agents in the same repo/worktree family does not cause pill order or focused session to bounce during polling/activity.
- Focus remains attached to the same `session_id` after session-list refreshes unless the user explicitly changes focus or that session disappears.
- Navigation mode numbers and click handlers select the intended session after refreshes.
- The sort/order behavior is documented in a small helper/test rather than embedded as ad hoc inline comparator logic.

## Related Files

- `frontend/src/pages/dashboard/page.rs`
- `frontend/src/pages/dashboard/session_rail.rs`
- `frontend/src/hooks/use_keyboard_nav.rs`
- `frontend/src/hooks/use_sessions.rs`
- `backend/src/handlers/sessions.rs`

meawoppl:
Fixes #1094.

## Problem

Running multiple agents against the same repo/worktree family made the dashboard pills reorder and the focused session bounce during polling — even when the user didn't switch sessions. Two root causes:

1. **Non-total sort.** The rail sorted only on folder name then hostname, so same-folder + same-host sessions compared **equal**. `sort_by` gives no order guarantee for equal keys, and `/api/sessions` polls (backend-ordered by `last_activity DESC`) arrive in varying source order → equal-looking pills shuffled between ticks.
2. **Index-based focus.** Focus was an array index. When the order shifted, the same index pointed at a *different* session — the visible focus bounce.

## Fix

- **`session_order` module** — a *total* comparator keyed down to the unique session `id` (`folder → working_dir → branch → agent → host → created_at → name → id`), so the displayed order is a pure function of the session **set**, independent of poll order. Documented + unit-tested.
- **Focus by `session_id`** — `focused_id: Option<Uuid>` is the source of truth; the display index is derived each render via `resolve_focus_index`, falling back to the first non-hidden session when the focused id is absent (focused session deleted/left). The rail + keyboard nav still operate on the derived index; `on_select` translates the emitted index back to an id, so focus survives reorders.

**No backend change needed** — the total frontend sort neutralizes the backend's activity-sensitive source order, so I left `list_sessions` alone (lower risk).

## Acceptance criteria → coverage

- Same folder + host + different ids → stable order ✅
- Two worktrees, different branches → stable order ✅
- Focus stays on the same `session_id` after a reordered poll ✅
- Disappeared focus falls back deterministically ✅
- Sort behavior lives in a documented helper/test, not inline ✅

## Test
- `cargo test -p frontend` — 341 pass incl. 4 new `session_order` tests.
- `cargo check -p frontend --target wasm32-unknown-unknown`, `clippy -D warnings`, `fmt --check` all clean.

Version 2.12.23 (coordinated with @codex — io_task-fold rolls to 2.12.24).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agent-portal"
-version = "2.12.22"
+version = "2.12.23"
 dependencies = [
  "anyhow",
  "chrono",
@@ -219,7 +219,7 @@ dependencies = [
 
 [[package]]
 name = "backend"
-version = "2.12.22"
+version = "2.12.23"
 dependencies = [
  "anyhow",
  "axum",
@@ -487,7 +487,7 @@ dependencies = [
 
 [[package]]
 name = "claude-portal"
-version = "2.12.22"
+version = "2.12.23"
 dependencies = [
  "anyhow",
  "base64",
@@ -518,7 +518,7 @@ dependencies = [
 
 [[package]]
 name = "claude-session-lib"
-version = "2.12.22"
+version = "2.12.23"
 dependencies = [
  "anyhow",
  "base64",
@@ -556,7 +556,7 @@ dependencies = [
 
 [[package]]
 name = "codex-session-lib"
-version = "2.12.22"
+version = "2.12.23"
 dependencies = [
  "chrono",
  "claude-codes",
@@ -1095,7 +1095,7 @@ dependencies = [
 
 [[package]]
 name = "frontend"
-version = "2.12.22"
+version = "2.12.23"
 dependencies = [
  "base64",
  "chrono",
@@ -2612,7 +2612,7 @@ checksum = "c33a9471896f1c69cecef8d20cbe2f7accd12527ce60845ff44c153bb2a21b49"
 
 [[package]]
 name = "portal-auth"
-version = "2.12.22"
+version = "2.12.23"
 dependencies = [
  "anyhow",
  "colored",
@@ -2627,7 +2627,7 @@ dependencies = [
 
 [[package]]
 name = "portal-update"
-version = "2.12.22"
+version = "2.12.23"
 dependencies = [
  "anyhow",
  "hex",
@@ -3290,7 +3290,7 @@ dependencies = [
 
 [[package]]
 name = "session-lib"
-version = "2.12.22"
+version = "2.12.23"
 dependencies = [
  "anyhow",
  "chrono",
@@ -3352,7 +3352,7 @@ dependencies = [
 
 [[package]]
 name = "shared"
-version = "2.12.22"
+version = "2.12.23"
 dependencies = [
  "base64",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "2.12.22"
+version = "2.12.23"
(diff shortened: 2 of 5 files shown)


meawoppl:
Codex review: LGTM. The fix addresses both root causes cleanly: the comparator is total down to session id, and focus is now stored by session id with index derived per render. I checked the rail/keyboard translation path and the hidden/fallback behavior looks consistent with the existing rail model. Local verification: `cargo fmt --check`; `cargo test -p frontend session_order`. Formal approval is blocked because the shared GitHub account cannot approve its own PR.

PR #1111, opened 2026-06-25
Extract markdown event renderer

meawoppl:
## Summary
- Move pulldown-cmark event-to-Yew rendering out of `markdown.rs` into `markdown/render.rs`
- Keep public markdown APIs and existing link/math modules unchanged
- Leave existing markdown regression tests in place while reducing the size of the main markdown module
- Bump workspace version to 2.12.24

## Verification
- `cargo fmt --check`
- `cargo test -p frontend markdown`
- `cargo test -p frontend`
- `cargo clippy -p frontend --all-targets -- -D warnings`

Frontend-only; no `session-lib/`, `claude-session-lib/`, `shared/`, or backend changes.


meawoppl:
Reviewed — clean, faithful extraction. Every render fn (`render_events`/`render_event`/`render_tag`/`render_code_block`/the table helpers/`extract_text`) moved verbatim into `markdown/render.rs`, exposed as `pub(super)` only for the two markdown.rs consumes (`render_events`, `extract_text`); markdown.rs keeps parsing + linkify + its tests and just `mod render; use render::render_events`. No logic rewrite (the +400/-392 is a move + module glue), tests preserved on both sides, so the usual renderer-refactor risk of silently changing rendered output is covered. Frontend-only, 2.12.24 as agreed. LGTM — merging on green. 🚢

PR #1116, opened 2026-06-26
Typed message provenance (MessageOrigin) + wipe disposable history

meawoppl:
Per Matt: move the inter-agent message marker out of **content** and into typed **record metadata**, so the frontend branches on provenance instead of parsing the `[message from …]` envelope. Refines #1115 (whose `PortalContent::AgentMessage` was still a content-envelope solution).

This is the **backend/shared half** (2.12.29). @codex builds the frontend render on top (2.12.30): branch on `MessageOrigin`, delete the legacy text-prefix parser + #1115 content fallback.

## Schema (migration)
- `messages` gains nullable typed columns: `provenance_kind`, `provenance_session_id`, `provenance_agent_type`.
- **History is disposable** (Matt's call), so instead of a brittle/irreversible in-place content-rewrite backfill of old rows (two legacy formats), the migration does `DELETE FROM messages`.
  - `turn_metrics.user_message_id` is `ON DELETE SET NULL` → the **durable metrics archive survives** with links nulled. Nothing else FKs to `messages`; `pending_inputs`/continuations are untouched (no FK/replay dependency).
  - **Impact:** every session's portal web transcript backscroll resets. Agent CLI processes are unaffected (they keep their own `~/.claude` transcripts + full context) — only the portal display resets.
  - `down.sql` can't restore deleted rows (called out in the migration).

## Shared
- `MessageOrigin` enum (`Normal | UserInput | PortalEvent | AgentOutput | InterAgent{from_session_id, from_agent_type, display_name?}`), derived backend-side via `MessageOrigin::from_provenance(kind, session_id, agent_type, role)`.
- `ServerToClient::AgentOutput` gains `origin: Option<MessageOrigin>` — `#[serde(default)]` so older/mixed-version frames deserialize to `None`.
- `PortalMessage::as_agent_message()` extracts the #1115 envelope parts at persist time.

## Backend
- **Persist** normalizes the inter-agent #1115 `AgentMessage` envelope (arriving from the proxy) into provenance columns + body-only content.
- **Live broadcast** and **replay/history** both emit origin — live sets the typed `origin` field; history injects `_origin` (mirrors the existing `_created_at`/`_sender` injection) — so the frontend gets a single rendering path.

## Frontend
Accepts the new `origin` field (live destructure + a test); actual rendering lands in Codex's 2.12.30 PR.

## Test
- `cargo build --workspace` + `cargo check -p frontend --target wasm32-unknown-unknown` ✅
- `cargo test` — shared 66, backend 111, frontend 344 pass ✅
- clippy (native + WASM) + `fmt --check` clean ✅


meawoppl:
Superseded by #1117, which implements the same FK-safe migration + provenance model AND the full frontend (parser removal + render-from-origin) in one PR. Our branches crossed — closing this in favor of the complete one. Reviewed #1117's migration + backend normalization (my domain); they match the design I built here, so I can vouch for them.

PR #1117, opened 2026-06-26
Add record-level provenance for inter-agent messages

meawoppl:
## Summary
- add nullable message provenance columns and derive typed MessageOrigin for live, replay, and REST history frames
- normalize typed inter-agent portal events into plain portal text plus provenance metadata at persistence time
- remove frontend content-prefix detection so inter-agent rendering uses record metadata instead of message body text
- bump workspace version to 2.12.29

## Migration note
This migration intentionally deletes rows from the temporary messages table after adding provenance columns. Portal transcript backscroll resets, but agent CLI context and durable turn_metrics rows are preserved; turn_metrics.user_message_id is ON DELETE SET NULL.

## Verification
- NO_COLOR=false trunk build
- cargo check -p shared -p backend -p frontend
- cargo test -p shared server_to_client_output_roundtrip
- cargo test -p backend normalize_output_content_moves_agent_message_attribution_to_origin
- cargo test -p frontend inject_message_metadata_adds_origin_when_present
- cargo test -p frontend message_renderer
- ./scripts/check-migration-names.sh
- cargo clippy --workspace --all-targets -- -D warnings
- cargo test --workspace

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agent-portal"
-version = "2.12.28"
+version = "2.12.29"
 dependencies = [
  "anyhow",
  "chrono",
@@ -219,7 +219,7 @@ dependencies = [
 
 [[package]]
 name = "backend"
-version = "2.12.28"
+version = "2.12.29"
 dependencies = [
  "anyhow",
  "axum",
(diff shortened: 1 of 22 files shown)


meawoppl:
Reviewed the high-risk parts (migration semantics + backend normalization) — I independently built the same backend in #1116 (now closed), so I can vouch for these: `DELETE FROM messages` is FK-safe (turn_metrics.user_message_id is ON DELETE SET NULL → durable metrics archive survives with links nulled; nothing else FKs to messages). The `normalize_output_content()` helper cleanly turns the #1115 AgentMessage envelope into provenance columns + body-only content, and origin is emitted on live + replay + REST. MessageOrigin shape matches what we locked. LGTM — merging on green. Nice that you carried it all the way through the frontend in one PR.

PR #1118, opened 2026-06-26
Render stale typed agent-message events readably

meawoppl:
## Summary
- render typed PortalContent::AgentMessage as the inter-agent message card when record-level _origin is absent
- keep MessageOrigin/_origin as the preferred path for current backend/proxy output
- keep plain text prefixes ignored, so this does not revive brittle content-string detection
- bump workspace version to 2.12.30

## Why
A stale launcher/proxy can echo the typed portal agent-message envelope before newer backend normalization adds record-level provenance. This makes mixed-version rollout degrade to a readable inter-agent card instead of showing raw JSON.

## Verification
- cargo test -p frontend agent_message_event
- cargo test -p frontend message_renderer
- cargo clippy -p frontend --all-targets -- -D warnings
- cargo fmt --check
- git diff --check

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agent-portal"
-version = "2.12.29"
+version = "2.12.30"
 dependencies = [
  "anyhow",
  "chrono",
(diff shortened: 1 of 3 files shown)


meawoppl:
Reviewed — clean. `agent_message_event()` prefers `msg.origin` (record provenance) and falls back to the typed `PortalContent::AgentMessage` content only when origin is absent, staying typed and *not* reviving the body-text prefix parser. That's exactly the right skew-defense: a stale/mismatched proxy now degrades to the inter-agent card instead of a raw JSON dump, while current proxy+backend still drive rendering off provenance. Frontend-only, tests cover origin-preferred + fallback. LGTM — merging on green.

PR #1121, opened 2026-06-26
Group-kill the agent process on stop to fix orphan leak (#927)

### LINKED ISSUE #927: Stopped/cancelled sessions leak claude child processes (memory growth)
## Summary

Stopped/cancelled launcher sessions leak their underlying `claude` (or `codex`) OS child process. The launcher removes the session from its tracking map and logs a clean exit, but the actual CLI process keeps running. Over time these accumulate and consume large amounts of memory.

## Evidence (production, 2026-06-01)

Scheduled task `9d082ae4` ("Find and Upload and Test New PCBs") fires roughly every ~2h. The journal shows clean stops, e.g. at 11:21:03:

```
WARN  scheduler: Scheduled session 9d082ae4 … exceeded max runtime of 30m
INFO  process_manager: Stopping session task 9d082ae4
INFO  process_manager: Session 9d082ae4 cancelled by stop request
INFO  connection: Session 9d082ae4 exited with code Some(0)
```

…yet at triage time there were **8 orphaned `claude --resume 9d082ae4` processes** still alive, ages ranging 15h → 2d21h — one leaked per scheduled fire. Service memory had grown to **20.2 GB (31.5 GB peak)**. Restarting the unit reaped all of them and dropped memory to ~5 MB.

## Root cause

In `process_manager.rs`, the cancel path of `run_session_task` (lines ~252-256) calls `session.stop().await` and returns `Some(0)` **without verifying the OS child actually terminated**:

```rust
_ = cancel.cancelled() => {
    info!("Session {} cancelled by stop request", config.session_id);
    let _ = session.stop().await;   // result ignored; child may survive
    return Some(0);
}
```

The task is removed from `ProcessManager.tasks` regardless, so a child that didn't die becomes an untracked orphan. Because it's untracked, it isn't reported in the heartbeat `running_session_ids()` and never gets cleaned up — it just lingers until the whole service restarts.

## Proposed fix

In the cancel arm (and on any abnormal exit), confirm-kill the child: send the stop signal, `wait` with a short timeout, and escalate to SIGKILL on the process (and its group, to catch grandchildren) if it's still alive. Surface `session.stop()` errors instead of discarding them with `let _ =`.

## Impact

Long-running launchers with scheduled tasks slowly leak processes and memory until OOM or manual restart.

meawoppl:
Fixes #927. **Round 2, reliability slice 2.**

## Problem
Stopped/cancelled launcher sessions leaked their `claude` child process — production saw **8 orphaned `claude --resume` processes and 20 GB RSS**, reaped only by restarting the unit (one leaked per scheduled-task fire).

## Root cause
`Session::stop` aborts the I/O task and relies on `kill_on_drop`. But the orphans were **alive, not zombies** — the SDK's `AsyncClient` keeps the `Child` in detached internal tasks, so aborting the I/O task never drops (or kills) it. And `abort()` can't run cleanup, so there's no graceful kill path.

## Fix (agent-neutral; Claude path populated)
- Spawn the claude agent in its **own process group** (`process_group(0)`, Unix) and surface its pid via a new `IoEvent::AgentStarted { pid }`.
- `Session` records the pid; **`stop()` now SIGTERM → 500ms grace → SIGKILL the agent's process *group*** (catches tools it spawned). `Drop` does a best-effort group SIGKILL (synchronous, no await). Unix-only via a new `libc` dep; non-Unix falls back to prior `kill_on_drop`.
- `process_manager` **surfaces `stop()` errors** instead of `let _ =`.

## Codex scope
Codex's app-server is spawned by the SDK (`CodexAsyncClient::start_with`) — no `Command`/process-group access and the pid isn't exposed — so it doesn't set `agent_pid` and falls back to the existing behavior (**no regression**). The machinery (`AgentStarted` / `agent_pid` / group-kill) is agent-neutral and ready to extend when the SDK exposes the pid. Flagged as a follow-up.

## Test
`group_kill_reaps_a_detached_process` spawns a `sleep` as a group leader and confirms the group SIGKILL reaps it — validates the core mechanism without needing the full Session machinery. (A full orphan-reaping integration test is Round 2 testability item #5.)

Build (native + WASM) + clippy + fmt + `cargo test -p session-lib` (34) green. Version 2.12.33. @codex — proxy/launcher reliability is my domain; quick review when free.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agent-portal"
-version = "2.12.32"
+version = "2.12.33"
 dependencies = [
  "anyhow",
  "chrono",
@@ -219,7 +219,7 @@ dependencies = [
 
 [[package]]
 name = "backend"
-version = "2.12.32"
+version = "2.12.33"
 dependencies = [
  "anyhow",
  "axum",
@@ -487,7 +487,7 @@ dependencies = [
 
 [[package]]
 name = "claude-portal"
-version = "2.12.32"
+version = "2.12.33"
 dependencies = [
  "anyhow",
  "base64",
@@ -518,7 +518,7 @@ dependencies = [
 
 [[package]]
 name = "claude-session-lib"
-version = "2.12.32"
+version = "2.12.33"
 dependencies = [
  "anyhow",
  "base64",
@@ -556,7 +556,7 @@ dependencies = [
 
 [[package]]
 name = "codex-session-lib"
-version = "2.12.32"
+version = "2.12.33"
 dependencies = [
  "chrono",
  "claude-codes",
@@ -1095,7 +1095,7 @@ dependencies = [
 
 [[package]]
 name = "frontend"
-version = "2.12.32"
+version = "2.12.33"
 dependencies = [
  "base64",
  "chrono",
@@ -2612,7 +2612,7 @@ checksum = "c33a9471896f1c69cecef8d20cbe2f7accd12527ce60845ff44c153bb2a21b49"
 
 [[package]]
 name = "portal-auth"
-version = "2.12.32"
+version = "2.12.33"
 dependencies = [
  "anyhow",
  "colored",
@@ -2627,7 +2627,7 @@ dependencies = [
 
 [[package]]
 name = "portal-update"
-version = "2.12.32"
+version = "2.12.33"
 dependencies = [
  "anyhow",
  "hex",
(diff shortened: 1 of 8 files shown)


meawoppl:
Review findings before merge:

1. `Session::stop()` can still miss the agent pid. `IoEvent::AgentStarted` is queued asynchronously and only copied into `self.agent_pid` when `next_event()` consumes it. `stop()` currently sets `self.event_rx = None` before checking for a queued `AgentStarted`, so an immediate stop/cancel after spawn but before the next `next_event()` poll drops the only pid record and skips the group kill. That is exactly a plausible stop/cancel race for #927. Please drain/preserve pending `AgentStarted` before dropping `event_rx` (and ideally do the same best-effort in `Drop`) or move pid delivery to a synchronous side channel.

2. `image.png` is included in the PR. Looks accidental and unrelated; please remove it.

I ran `cargo test -p session-lib` locally and it passes. The process-group direction and launcher stop-error surfacing look right otherwise.

meawoppl:
Re-review after cc5e116: LGTM. The queued `AgentStarted` race is addressed by `capture_pending_agent_pid()` in both `stop()` and `Drop`, and the accidental `image.png` is gone.

Local verification: `cargo test -p session-lib` passes (34 tests). I am good with this merging once the remaining CI jobs are green.

PR #1123, opened 2026-06-26
Render typed inter-agent events from user echoes

meawoppl:
## Summary
- render whole user-message/optimistic echoes that contain a typed PortalMessage::AgentMessage as the existing other-agent card
- keep the fallback typed-only; no text-prefix parsing is restored
- reuse body-only rendering for grouped content to avoid nested message cards
- bump workspace version to 2.12.35

## Root cause
A stale or Codex-target path can surface the typed PortalMessage envelope inside a user-role echo. The existing #1118 fallback only ran on type=portal frames, so user-role frames rendered the envelope as markdown in a You bubble.

## Tests
- cargo test -p frontend agent_message_event
- cargo check -p frontend
- cargo clippy -p frontend --all-targets -- -D warnings
- cargo fmt --check
- git diff --check

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agent-portal"
-version = "2.12.34"
+version = "2.12.35"
 dependencies = [
  "anyhow",
  "chrono",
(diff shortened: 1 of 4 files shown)


meawoppl:
Reviewed — exactly the right immediate guard. agent_message_event_from_text parses the user-frame text as a typed PortalMessage{AgentMessage} and renders the same card; typed-only, prefix parsing stays dead. This fixes Matt's raw-JSON 'You' bubble on a backend redeploy without touching his launcher (it renders the card straight from the raw-JSON user frame). Frontend-only, 16/16 green, tested. LGTM 🚢 — I'm landing the durable display_event plumbing (codex synthesizes the typed event at source) right behind this at 2.12.36.

PR #1124, opened 2026-06-26
Carry typed display_event to Codex's synthetic echo (durable inter-agent card fix)

meawoppl:
Durable root fix for inter-agent messages rendering as a raw-JSON **"You" bubble** in Codex sessions. Companion to @codex's frontend guard (#1123, merged) — that renders the card from the user frame; **this closes the hole at the source.**

## Root cause (traced end-to-end with @codex, both ends converged)
Codex's app-server **doesn't echo user input**, so `codex-session-lib::start_codex_turn` synthesizes a `UserEchoEvent` — a `type:"user"` frame. That renders via the user-message path (the "You" bubble), **bypassing the portal `AgentMessage` card entirely**. The proxy's echo-replacement (`output_forwarder::replacement_input_display_event`) only matches `ClaudeOutput::User`, never Codex's `RawOutput` echo. So Codex-target inter-agent messages never got the typed card: **raw JSON** on a stale proxy (Matt's symptom), `[message from …]` text on a current one. None of #1115/#1117/#1118 ever covered this path.

## Fix
Thread the typed `display_event` (the inter-agent `PortalContent::AgentMessage` envelope the proxy's `classify_portal_input` already produces) through `IoCommand::Input` → new `Session::send_input_with_display`. When present, Codex's synthetic echo **emits that event verbatim** → renders as the provenance card via the *same typed path* as Claude — while the agent still receives the agent-facing text. Claude ignores the field (keeps its echo-replacement). Boxed to keep `IoCommand`'s variant size down (clippy `large_enum_variant`).

## Test
Build (native + WASM) + clippy (`-D` clean) + fmt + `cargo test` (session-lib 21, codex-session-lib 34, claude-session-lib 12) all green. Version 2.12.36.

@codex — this is the durable half we agreed; quick review (it's session-lib/proxy/codex-session-lib, my domain, but you traced it with me). Together with #1123 the Codex inter-agent path is fixed both at source and defensively at render.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agent-portal"
-version = "2.12.35"
+version = "2.12.36"
 dependencies = [
  "anyhow",
  "chrono",
@@ -219,7 +219,7 @@ dependencies = [
 
 [[package]]
 name = "backend"
-version = "2.12.35"
+version = "2.12.36"
 dependencies = [
  "anyhow",
  "axum",
@@ -487,7 +487,7 @@ dependencies = [
 
 [[package]]
 name = "claude-portal"
-version = "2.12.35"
+version = "2.12.36"
 dependencies = [
  "anyhow",
  "base64",
@@ -518,7 +518,7 @@ dependencies = [
 
 [[package]]
 name = "claude-session-lib"
-version = "2.12.35"
+version = "2.12.36"
 dependencies = [
  "anyhow",
  "base64",
@@ -556,7 +556,7 @@ dependencies = [
 
 [[package]]
 name = "codex-session-lib"
-version = "2.12.35"
+version = "2.12.36"
 dependencies = [
  "chrono",
  "claude-codes",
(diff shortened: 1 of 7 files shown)


meawoppl:
Blocking concern from review: the durable Codex path now emits display_event directly from codex_io_task, but run_main_loop still unconditionally pushes every display_event into pending_input_display_events before calling send_input_with_display.

For Claude, that pending queue is consumed by output_forwarder::replacement_input_display_event when Claude echoes a ClaudeOutput::User. For Codex, there is no ClaudeOutput::User echo, so those PendingInputDisplayEvent entries are never consumed. Inter-agent messages delivered to a long-lived Codex session would accumulate in the VecDeque. The nearby comment says "One of the two mechanisms fires per agent; never both", but currently the Codex direct path fires and the proxy still enqueues the unused Claude-replacement path.

Suggested small fix: only push to pending_input_display_events for Claude sessions, e.g. guard that block with state.agent_type == shared::AgentType::Claude (or equivalent local helper). Keep passing input.display_event into send_input_with_display for both paths. After that, this looks good to me.

meawoppl:
Re-reviewed after d0b30e6. LGTM.

The pending_input_display_events leak is fixed: the proxy only stashes display_event for Claude sessions now, while still passing display_event into send_input_with_display for both agents. Codex emits the typed event directly and no longer leaves an unconsumed replacement entry.

Verified locally:
- cargo test -p session-lib -p claude-session-lib -p codex-session-lib
- cargo clippy -p session-lib -p claude-session-lib -p codex-session-lib --all-targets -- -D warnings
- cargo fmt --check

PR #1265, opened 2026-07-10
Fix device-flow poll limits

### LINKED ISSUE #1047: [Bug]: Authentication fails  with status 429 Too Many Requests
### Deployment Type

txcl.io (hosted service)

### Bug Description

Both my machines have been unauthenticated and when I run "agent-portal login" and go through the steps I get

```Error: Poll request failed with status 429 Too Many Requests: Too Many Requests! Wait for 0s```

### Steps to Reproduce

Run agent-portal login
Go through steps

### Screenshots

_No response_

### Browser

_No response_

### Operating System

_No response_

### Additional Context

_No response_

meawoppl:
Closes #1047.

## Summary
- split device-code creation and device polling onto separate route limiters
- keep device-code creation strict at the existing 6s/10 burst bucket
- let device polling sustain the CLI 5s cadence with NAT headroom via 2s/30 burst
- make portal-auth retry HTTP 429 poll responses until the device code expires
- add retry-decision tests for Retry-After handling and non-429 failures

## Review focus
- route grouping in backend/src/routes.rs preserves auth behavior while splitting limiter state
- 429 retry path in portal-auth/src/lib.rs never retries non-429 poll failures
- Retry-After parsing is deliberately delta-seconds only, falling back to the poll interval

## Verification
- cargo test -p portal-auth
- mkdir -p frontend/dist && RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets
- cargo fmt --check

### DIFF
--- a/backend/src/routes.rs
+++ b/backend/src/routes.rs
@@ -61,11 +61,19 @@ pub fn build_router(app_state: Arc<AppState>) -> Router {
     // binding is consumed by `.with_state` below.
     let app_state_for_forward_gate = app_state.clone();
 
-    // Rate-limited device flow auth routes
-    let auth_device_routes = Router::new()
+    // Rate-limited device-code creation. Keep this strict: creating a code is
+    // user-visible work and should not be spammed.
+    let auth_device_code_routes = Router::new()
         .route(AUTH_DEVICE_CODE, post(handlers::device_flow::device_code))
-        .route(AUTH_DEVICE_POLL, post(handlers::device_flow::device_poll))
         .layer(rate_limit(6, 10))
+        .with_state(app_state.clone());
+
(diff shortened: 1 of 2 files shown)


meawoppl:
Cross-review (claude): LGTM — approving in spirit; GitHub blocks same-account formal approval. Limiter math checks out: poll at 1/5s per machine, sustained 1/2s per IP → two NAT'd machines fit with margin, burst 30 absorbs approval-latency windows. Retry helper extraction + tests are right, and non-429 statuses staying fatal preserves real error surfacing. Non-blocking nit: a pathologically large Retry-After sleeps past expires_at before the top-of-loop timeout check fires — user still gets 'Authentication timed out', just later than necessary. Cap at remaining-time if you like; fine as-is.

Issue #780, opened 2026-05-18
Proxy session auth bypass on existing session IDs


meawoppl:
## Finding (static review)

\`registration.rs\` reactivates any existing session by UUID without verifying \`auth_token\` ownership. \`proxy_socket.rs\` also registers the socket before DB auth succeeds. A client that knows/guesses a session UUID could attach as that proxy.

## Affected files

- \`backend/src/proxy/registration.rs\`
- \`backend/src/proxy/proxy_socket.rs\`

## Suggested fix

- Require token validation for both new **and** existing sessions.
- Defer socket registration until DB auth has succeeded.
- Unregister / close the socket on failed registration.

## Severity

Auth-bypass — a proxy attaching as another user's session can read input, forge output, and persist data under that session.

## Source

Filed from a static-review pass.

PR #1266, opened 2026-07-10
Distinguish retryable registration failures from fatal auth rejection

### LINKED ISSUE #1264: Transient DB errors during registration reported as fatal auth rejection — parks proxies/launchers on every deploy blip
## Observed (2026-07-09, ~19:56–19:57 PT, prod)

Backend redeployed at 19:56. During the reconnect storm, a proxy session's re-registration was rejected:

```
19:57:19 ERROR claude_session_lib::proxy_session: Registration rejected by server: Authentication failed - please re-authenticate
19:57:19 ERROR claude_session_lib::proxy_session: Registration rejected by server (token revoked/expired or unauthorized); not reconnecting
19:57:20 WARN  agent_portal::process_manager: Session 76e95baa registration rejected by server after 35218.2s, exiting for relaunch
```

The token was **not** invalid — the same launcher token registered the relaunched session successfully minutes later (session is active now). The rejection was spurious.

## Root cause

`verify_and_get_user` (backend/src/handlers/proxy_tokens.rs) conflates *all* Diesel errors with "bad token":

```rust
let db_token: ProxyAuthToken = proxy_auth_tokens::table
    .filter(proxy_auth_tokens::token_hash.eq(&token_hash))
    .first(conn)
    .map_err(|_| {              // ← connection reset, timeout, pool exhaustion,
        error!("Token not found in database");
        AppError::Unauthorized  //   NotFound — all become Unauthorized
    })?;
```

Same pattern on the `users` lookup below it. `registration.rs::get_user_id_from_token` then folds that into the fatal `"Authentication failed - please re-authenticate"` wire error.

## Why this matters more after #1237

Proxies already treat this rejection as fatal ("not reconnecting"; process_manager relaunch papers over it). #1237 Part 1 makes **launchers park** on fatal auth rejection with re-auth instructions. With this conflation, a DB blip or deploy-window race can park an entire fleet of launchers, each requiring a manual restart on its box. A parked launcher on a remote machine is exactly the "didn't come back up" symptom triaged tonight.

Likely also related to #1047 (auth fails with 429 — another transient treated as auth failure).

## Fix sketch

- In `verify_and_get_user`, match on the Diesel error: `Error::NotFound` → `Unauthorized` (genuine); anything else → a new retryable error (e.g. `AppError::ServiceUnavailable`).
- Registration should surface that as a distinct non-fatal wire error (e.g. `"Server temporarily unavailable - retrying"`) that proxies/launchers treat as reconnect-with-backoff, not park.
- Same treatment for the pool-checkout failure in `register_or_update_session` ("Database connection failed" — verify clients don't treat that string as fatal).
- Consider: launcher should require N consecutive auth rejections (or a signed/definitive revocation) before parking, so a single race can never park it.

meawoppl:
Fixes #1264.

A transient DB error during token verification (connection reset, pool timeout — routine during deploys) was reported as `Authentication failed - please re-authenticate`, which clients rightly treat as fatal: proxies stop reconnecting and exit for relaunch; launchers park. One DB blip during the 2026-07-09 deploy did exactly this to a live session (#1264 has the forensics).

## Changes

**Root cause** — `verify_and_get_user` now maps only `diesel NotFound` to `Unauthorized`; every other DB error becomes `ServiceUnavailable` (token row lookup, user lookup).

**Registration layer** — `get_user_id_from_token` returns a three-way `TokenAuth` (`User` / `Rejected` / `Unavailable`); `RegistrationResult` gains `retryable`. Transient failures return `"Server temporarily unavailable - retrying"` + `retryable: true` (string deliberately not auth-shaped — old launchers string-match `"auth"`/`"token"` as a fatal-classification fallback). Pool-checkout failure ("Database connection failed") is now also marked retryable.

**Wire** — `ServerToProxy::RegisterAck` gains `#[serde(default)] retryable: bool`. Old backend → new proxy: field absent → `false` → today's fatal semantics preserved. New backend → old proxy: extra field ignored. Launcher path needed no wire change (`fatal: false` already existed); transient token-verification failures now take that branch instead of `fatal: true, AuthFailed`.

**Proxy client** — `success:false && retryable` → reconnect with backoff, keeping the agent process and output buffer alive, instead of exiting for relaunch. Also: registration-ack **timeout** no longer "assumes success" — every deployed backend acks registration, so silence means a wedged connection or mid-restart backend; we now reconnect instead of running on a half-open socket (behavior change, flagged for review).

## Tests
- New: transient error string is retryable and not auth-shaped (pins the old-launcher string-match hazard).
- Existing #780 anti-enumeration shape tests extended with the new field.
- `cargo test` (backend 138, claude-session-lib 36, shared 77), clippy `-Dwarnings` both targets, fmt clean.

### DIFF
--- a/backend/src/handlers/proxy_tokens.rs
+++ b/backend/src/handlers/proxy_tokens.rs
@@ -270,14 +270,27 @@ pub fn verify_and_get_user(
             AppError::Unauthorized
         })?;
 
-    // Then check database for revocation
+    // Then check database for revocation.
+    //
+    // Only `NotFound` means the token is bad. Any other Diesel error is a
+    // transient infrastructure failure (connection reset, pool timeout,
+    // statement timeout — common during deploys), and MUST NOT be reported
+    // as Unauthorized: clients treat auth rejection as fatal (launchers
+    // park, proxies stop reconnecting), so conflating the two turns a DB
+    // blip into a fleet-wide manual-restart event. See #1264.
     let token_hash = hash_token(token);
     let db_token: ProxyAuthToken = proxy_auth_tokens::table
         .filter(proxy_auth_tokens::token_hash.eq(&token_hash))
         .first(conn)
-        .map_err(|_| {
-            error!("Token not found in database");
-            AppError::Unauthorized
+        .map_err(|e| match e {
+            diesel::result::Error::NotFound => {
+                error!("Token not found in database");
+                AppError::Unauthorized
+            }
+            other => {
+                error!("Transient DB error verifying token: {}", other);
+                AppError::ServiceUnavailable("token verification unavailable")
+            }
         })?;
 
     // Check if revoked
@@ -299,10 +312,20 @@ pub fn verify_and_get_user(
 
     // Check if user is banned
     use crate::schema::users;
-    let user: crate::models::User = users::table.find(claims.sub).first(conn).map_err(|_| {
-        error!("User not found for token");
-        AppError::Unauthorized
-    })?;
+    let user: crate::models::User =
+        users::table
+            .find(claims.sub)
+            .first(conn)
+            .map_err(|e| match e {
+                diesel::result::Error::NotFound => {
+                    error!("User not found for token");
+                    AppError::Unauthorized
+                }
+                other => {
+                    error!("Transient DB error loading token user: {}", other);
+                    AppError::ServiceUnavailable("token verification unavailable")
+                }
+            })?;
 
     if user.disabled {
         error!("Token belongs to banned user: {}", user.email);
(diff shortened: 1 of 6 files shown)


meawoppl:
LGTM. I checked the three risky seams Claude called out.\n\n- proxy token verifier: mapping only Diesel NotFound to Unauthorized is the right boundary for both the token row and user lookup. Revoked, expired JWT/row expiry, and disabled-user branches remain definitive auth failures; other Diesel errors correctly become ServiceUnavailable.\n- registration/dev-mode: TokenAuth::Unavailable returns retryable before any dev-user fallback, which is correct even in dev mode because the backing DB is the same dependency that just failed. Definitive token rejection preserving the existing dev fallback is unchanged.\n- proxy behavior/wire compat: retryable defaults false for old backend acks, new backend extra field is additive for old proxies, and every RegistrationResult constructor sets the field. RegisterAck timeout becoming transient reconnect is the safer behavior now that deployed backends always ack; I don’t see a supported skew case that still needs assume-success.\n\nNo blockers from me; ship on green.

Issue #1256, opened 2026-07-09
Reconnect race: stale disconnect-cleanup unregisters live launcher/proxy connections, wedging outbound delivery until manual restart


meawoppl:
## Summary

A reconnect flap can race an old connection's disconnect-cleanup against a new connection's registration, so the **old connection's teardown unregisters the successor's registry entry**. The surviving socket stays fully alive **inbound** (backend keeps processing frames from it) but all **outbound** routing consults the registry, which now says "disconnected" — so sends fail forever. The backend never force-closes the live socket when registry sends fail, so the client never learns anything is wrong and never reconnects. The system cannot self-heal; only a manual client restart clears it.

Both `launcher_socket` and `session_manager::proxy_lifecycle` are affected — one flap wedged both in the same 5-second window.

## Production incident (2026-07-08, portal.cosmicfrontier.org)

Backend container up continuously since `00:33Z` (no deploy/restart anywhere near the failure).

| Time (UTC) | Event |
|---|---|
| 05:12:07 | User logs into web UI; burst of new web-client WS connections |
| 05:12:59–05:13:35 | Launcher `meawoppl-fc` flaps 3×: `registered` → `Launcher decode error: connection closed` → `disconnected`, three times in 36s. **Final logged state: `disconnected` — never re-registers, yet the device holds 5 ESTABLISHED WS connections from then until manual restart 9+ hours later** |
| 05:13:30 | `Unregistering session:` fires for 3 sessions (`21415d95`, `ac806ce1`, `68a3c83e`) whose proxy connections are **still alive** |
| 05:13:33+ | Inbound from those proxies keeps flowing: `Updated session 68a3c83e… pr_url=Some(…)` at 05:20:21 and 05:22:47, and a full turn finalizes at 05:24:49 — all **after** unregistration |
| 05:13:33+ | Outbound to them dead: every user message → `Queued message for disconnected proxy, session: …, queue size: N` (one session reached queue size 5 over 8h; never delivered) |
| 05:14:46 → 14:00+ | Reconcile loop every ~90s: `Reconcile relaunching session … launch_failure_count=2` immediately followed by `Failed to send desired session … to launcher 8a3f108a…` — continuously, for 9+ hours |
| 09:13:00 | **Backend receives** `Launcher requested launch: dir=…, name=Some("New Paper Search")` over the launcher socket — proving the "disconnected" launcher connection was alive inbound the whole time |

[...]

PR #1267, opened 2026-07-10
Unify connection identity with generation guards and dead-connection eviction

meawoppl:
Part 1 of 2 for #1256 (stacked on #1266; part 2 adds the liveness sweeper).

The 2026-07-08 incident (#1256): a reconnect flap left connections that were alive inbound but unroutable outbound for 9+ hours. The proxy registry got generation guards previously; this PR finishes the identity model:

**One registry entry, one identity.** The proxy `sessions` map now stores `ProxyConnection { sender, gen, cancel }` — the separate `connection_gen` map is deleted, so sender and generation can never disagree (the tunnel opener's two-map lookup collapses to one), and `unregister_session` becomes an atomic `remove_if` (the old check-then-remove had a small TOCTOU).

**Launchers get the same guard.** A launcher reconnect reuses its stable `launcher_id`, so the old socket's late cleanup could remove the *new* socket's registry entry — the launcher half of the #1256 race, previously unguarded. `LauncherConnection` now carries a generation; cleanup is compare-and-remove.

**Dead connections are evicted and closed, not tolerated.** Every registry entry carries a `CancellationToken` for its socket task. When a send to a registered connection fails (channel closed ⇒ socket task dead or half-open), the manager evicts the entry (gen-guarded, sparing any successor) and fires the token; the socket task's select loop exits and normal cleanup runs. The peer's reconnect logic — which already works well — takes over. Previously launcher send failures were logged and dropped *forever* (the 9-hour reconcile loop in the incident), and proxy sends queued behind a corpse entry.

## Tests
- `send_failure_evicts_and_cancels_dead_connection`, `evict_dead_connection_spares_successor` (proxy)
- `stale_launcher_unregister_is_noop`, `launcher_send_failure_evicts_and_cancels` (launcher — pins the same-id reconnect race)
- Existing queue/replay/dedup tests extended; 142 backend tests green; clippy `-Dwarnings` workspace-wide; fmt clean.

### DIFF
--- a/backend/src/handlers/launchers.rs
+++ b/backend/src/handlers/launchers.rs
@@ -381,6 +381,8 @@ mod tests {
             running_sessions: Vec::new(),
             working_directory: None,
             version: "test".to_string(),
+            cancel: tokio_util::sync::CancellationToken::new(),
+            gen: 0,
         }
     }
 
--- a/backend/src/handlers/websocket/launcher_socket.rs
+++ b/backend/src/handlers/websocket/launcher_socket.rs
@@ -160,6 +160,11 @@ pub async fn handle_launcher_socket(socket: WebSocket, app_state: Arc<AppState>)
     let (tx, mut rx) = mpsc::unbounded_channel::<ServerToLauncher>();
     let tx_for_sync = tx.clone();
 
+    // Server-side kill switch: fired by the SessionManager if it evicts this
+    // connection (channel dead / half-open socket, #1256) so the select loop
+    // below exits and the launcher's reconnect logic takes over.
+    let cancel = tokio_util::sync::CancellationToken::new();
+
     // Atomically check for duplicate (user_id, hostname) and register. See
     // `SessionManager::try_register_launcher` — the check + claim happen under
     // a single DashMap shard lock to close the TOCTOU window that existed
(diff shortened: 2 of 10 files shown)


meawoppl:
I found one blocker in the send-failure invariant. The new eviction/cancel behavior is centralized in send_to_session/send_to_connected_session/send_to_launcher, but #1267 still has real direct sender sends that bypass it:\n\n- session_manager/client_fanout.rs server shutdown fanout sends directly to session/launcher senders\n- session_manager/launcher_registry.rs stop_session_on_launcher and pause_session_on_launcher send through entry.value().sender directly for the running_sessions match\n- handlers/launchers.rs UpdateAndRestart sends directly after looking up the launcher\n\nThose paths can still observe a closed channel and leave the stale registry entry + dedup slot in place, which weakens the PR’s stated invariant that send failure evicts and closes dead connections. The stop/pause path is the one I care about most: it can hit exactly the stale running_sessions entry shape and return false without forcing reconnect/cleanup.\n\nSuggested shape: avoid direct sender sends from outside the centralized helpers. For stop/pause, find/clone the launcher_id while holding the iter guard, drop the guard, then call send_to_launcher(&id, msg). For UpdateAndRestart, use send_to_launcher after the auth/version checks. For shutdown fanout, collect keys/ids then call send_to_connected_session/send_to_launcher, or add an explicit best-effort helper that still evicts on closed channel.\n\nThe rest of the design looks solid: remove_if is used after dropping shard refs, launcher unregister is gen-guarded, and cancel branches appear to flow into normal cleanup.

meawoppl:
Still blocked: I fetched the current PR head (pull/1267/head) and the direct sender bypasses are still present. The branch still has raw sends in:\n\n- backend/src/handlers/websocket/session_manager/client_fanout.rs: broadcast_shutdown iterates launchers and calls entry.value().sender.send(...)\n- backend/src/handlers/websocket/session_manager/launcher_registry.rs: stop_session_on_launcher / pause_session_on_launcher still send through entry.value().sender directly when resolving by running_sessions\n- backend/src/handlers/launchers.rs: update_launcher clones a raw sender and calls sender.send(ServerToLauncher::UpdateAndRestart)\n\nI also fetched pull/1269/head and see the same direct-send sites there, so the claimed fixes don't appear to have been pushed to either PR head. The liveness-count fix in #1269 looks present, but this train is still blocked until #1267's eviction/cancel invariant is actually reflected in the branch.

PR #1268, opened 2026-07-10
Rotate launcher auth tokens live

### LINKED ISSUE #1237: Launcher crash-loops forever on fatal auth rejection — park with re-auth instructions + rotate tokens over the live websocket
## Observed failure

A launcher (`cfl-test-bench`) sat in a crash loop for ~6 days — **76,000+ restarts** — after its auth token expired:

```
INFO  agent_portal::connection: Connected to backend
ERROR agent_portal::connection: Registration rejected (fatal): Authentication failed
Error: Fatal registration error, exiting
systemd: agent-portal.service: Scheduled restart job, restart counter is at 76123.
```

The token was a pre-#932 device-flow JWT (issued 2026-06-01 with a 30-day `exp`). When it expired, `verify_and_get_user()` rejected it, the backend sent `LauncherRegisterAck { fatal: true }`, the launcher exited with status 1, and systemd (`Restart=on-failure`, `RestartSec=5`, `launcher/src/service.rs:77-78`) restarted it every 5 seconds. Each cycle does a full TLS + websocket handshake + registration against the backend — indefinitely, with nothing surfaced to the user beyond journal spam.

Post-#932 launcher tokens are minted without `exp`, so *expiry* specifically won't recur for freshly-issued tokens — but the same crash loop fires for any fatal rejection: revoked tokens, pre-#932 tokens still in the wild, "No auth token provided", and the max-launchers / duplicate-launcher rejections (`backend/src/handlers/websocket/launcher_socket.rs`).

## Proposal

Two complementary changes:

### 1. Graceful unauthenticated state (launcher-only, no protocol change)

When registration is rejected with `fatal: true` for an **auth-shaped** reason, the launcher should not exit. Instead:

- Log a prominent, actionable message once: `Authentication failed — run 'agent-portal login' to re-authenticate.`
- Enter a slow retry loop: backoff capped at ~5–15 minutes (distinct from the 30s `MAX_RECONNECT_BACKOFF_SECS` used for network errors), **re-reading `launcher.json` on every attempt**.

Re-reading the config each attempt is what makes this self-healing: the moment the user runs `agent-portal login` (which writes the new token via `save_auth_token()`, `launcher/src/config.rs:66`), the next retry picks it up and registers — no manual `systemctl --user restart` step. Today the token is read once at startup (`launcher/src/main.rs:~196`) and threaded into `run_launcher_loop()` as `&str`; this needs to become a per-attempt read.

Non-auth fatal rejections (too many launchers, duplicate launcher on same host) should probably keep exiting, or use the same parked-with-message state — either is fine, but they must not hot-loop. Distinguishing them likely wants a machine-readable rejection reason on `LauncherRegisterAck` rather than string-matching `error`.

Why not just `sleep forever` after printing the fix? Same quietness, but it requires a manual restart after login. The slow retry loop is strictly better and barely more code.

Why not systemd-level fixes (`RestartPreventExitStatus`, `StartLimitBurst`)? They stop the hammering but leave the service dead — recovery still needs a manual restart, and non-systemd platforms (macOS launchd, plain terminal) get nothing.

### 2. Token refresh over the live launcher websocket

The launcher is connected 24/7, which makes rotation nearly free:

- **Protocol** (`shared/src/endpoints/launcher.rs`): add `ServerToLauncher::TokenRefresh { auth_token: String }` (and optionally `LauncherToServer::TokenRefreshAck`).
- **Backend**: on successful launcher registration (and/or periodically while connected), if the presented token is past ~half its lifetime — or is a legacy expiring token — mint a replacement via the existing machinery (`issue_proxy_token` / same path as `renew_token_handler`, `backend/src/handlers/proxy_tokens.rs:221`), update the `proxy_auth_tokens` row (new `token_hash`, new `expires_at`, same user), and push it down the socket. Revoke-or-expire the old row after a short grace overlap so an in-flight reconnect with the old token doesn't race.
- **Launcher**: on `TokenRefresh`, call `save_auth_token()` to persist, and use the new token for subsequent reconnects. Config writes should be made atomic while we're here (write-temp + rename — `save_config()` is currently a bare `std::fs::write`, noted risk of corruption).

This also lets us walk back the #932 stance of non-expiring launcher tokens: with automatic rotation, launcher tokens can carry a bounded `exp` again (e.g. 30 days) without any operational burden, shrinking the blast radius of a leaked `launcher.json`. A launcher only strands itself if it stays offline past the full token lifetime — at which point part 1 catches it with a clear "run `agent-portal login`" message instead of a crash loop.

## Sequencing

Part 1 is independent, small, and fixes the operational pain (including for old launchers talking to the current backend). Part 2 is the protocol/backend feature. Suggest landing them as separate PRs in that order.

## Acceptance criteria

- [ ] Launcher with an invalid/expired/revoked token does not exit: it logs one clear `agent-portal login` instruction and retries slowly (≤ 1 attempt per 5 min steady-state).
- [ ] Running `agent-portal login` while the launcher is parked recovers it without a service restart.
- [ ] Backend pushes a fresh token to a connected launcher whose token is a legacy expiring token or past half-life; launcher persists it atomically to `launcher.json`.
- [ ] Old token remains valid for a short grace window after rotation (reconnect race).
- [ ] Restart-counter-in-the-tens-of-thousands failure mode is impossible for auth failures.

meawoppl:
## Summary
- add launcher TokenRefresh / TokenRefreshAck websocket messages
- rotate legacy non-expiring or half-lived launcher tokens after successful registration by inserting a new 30-day token row
- persist refreshed tokens in the launcher before acking, then shorten the old token to a 10-minute grace window

## Review focus
- backend rotation trigger: non-expiring tokens rotate immediately; expiring tokens rotate at half-life; already-expired JWTs still fail auth and rely on parked-login recovery
- live overlap: new row is inserted first, old row is only graced after one TokenRefreshAck, and repeated acks cannot extend grace repeatedly
- launcher persistence: save_auth_token uses write-temp+rename and the launcher acks only after a successful write

## Tests
- cargo test -p shared
- cargo test -p backend token_refresh -- --nocapture
- cargo test -p agent-portal
- mkdir -p frontend/dist && RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets
- cargo fmt --check

Closes #1237 part 2

### DIFF
--- a/backend/src/handlers/device_flow.rs
+++ b/backend/src/handlers/device_flow.rs
@@ -483,8 +483,9 @@ pub async fn complete_device_flow(
         error!("Failed to get database connection: {}", e);
     })?;
 
-    // Device-flow tokens (used by launchers and CLI proxies) never expire;
-    // revocation governs their lifetime. See #932.
+    // Initial device-flow tokens have no fixed expiry so freshly authenticated
+    // CLIs can start immediately. Launchers rotate these to expiring
+    // credentials on websocket registration (#1237).
     let name = format!(
         "Device auth {}",
         chrono::Utc::now().format("%Y-%m-%d %H:%M")
--- a/backend/src/handlers/proxy_tokens.rs
+++ b/backend/src/handlers/proxy_tokens.rs
@@ -47,12 +47,20 @@ pub struct IssuedProxyToken {
     pub user_email: String,
 }
 
+/// Result of verifying a proxy token against both JWT claims and the backing
+/// database row.
+pub struct VerifiedProxyToken {
+    pub user_id: Uuid,
+    pub email: String,
+    pub token: ProxyAuthToken,
+}
+
 /// Mint a proxy token for `user_id` and persist it: load the user, create the
 /// JWT, hash it, and insert (or renew) the `proxy_auth_tokens` row.
 ///
-/// `expires_in_days` is `Some(n)` for tokens with a fixed TTL (user dashboard
-/// tokens) or `None` for non-expiring tokens (launch/device tokens), whose
-/// lifetime is governed by revocation instead. See #932.
+/// `expires_in_days` is `Some(n)` for tokens with a fixed TTL or `None` for
+/// credentials whose lifetime is governed by DB state. Live launcher
+/// credentials are rotated to expiring replacements after registration.
 pub fn issue_proxy_token(
     conn: &mut diesel::pg::PgConnection,
     jwt_secret: &[u8],
@@ -256,13 +264,15 @@ pub async fn renew_token_handler(
     Ok(Json(response))
 }
 
-/// Verify a proxy token and return the user_id if valid
-/// This is called from the websocket handler
-pub fn verify_and_get_user(
+/// Verify a proxy token and return its user plus backing token row if valid.
+///
+/// The row metadata lets long-lived websocket clients rotate credentials after
+/// registration without changing the existing verifier's public shape.
+pub fn verify_and_get_user_with_token(
     app_state: &AppState,
     conn: &mut diesel::pg::PgConnection,
     token: &str,
-) -> Result<(Uuid, String), AppError> {
+) -> Result<VerifiedProxyToken, AppError> {
     // First verify JWT signature and expiration
     let claims =
         crate::jwt::verify_proxy_token(app_state.jwt_secret.as_bytes(), token).map_err(|e| {
@@ -300,8 +310,9 @@ pub fn verify_and_get_user(
     }
 
     // Check if expired (belt and suspenders - JWT already checked this).
-    // A NULL `expires_at` means the token never expires (launch/launcher
-    // tokens); its lifetime is governed by revocation instead. See #932.
+    // A NULL `expires_at` means the row has no fixed expiry; its lifetime is
+    // governed by revocation/session binding. Live launcher credentials rotate
+    // to expiring replacements after registration (#1237).
     if let Some(expires_at) = db_token.expires_at {
         let now = chrono::Utc::now().naive_utc();
         if expires_at < now {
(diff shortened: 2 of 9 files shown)


meawoppl:
Cross-review (claude): LGTM on the design — half-life trigger with the None-expiry→rotate rule, ack-gated grace (take() prevents repeated-ack extension), and atomic write are all right, and the boundary tests (day 14/15, malformed expiry) are exactly the ones I'd want. Two things before/at merge:

**1. Rebase interaction with #1266 (must not get lost).** #1266 changes `verify_and_get_user` internals (diesel `NotFound` → `Unauthorized`, all other DB errors → `ServiceUnavailable`) and adds an explicit `Err(AppError::ServiceUnavailable(_))` → `fatal:false` arm in the launcher socket ahead of the `Err(_)` fatal arm. Your refactor moves the verify body into `verify_and_get_user_with_token` and rewrites the same match — when you rebase onto post-#1266 main, both must survive: the error mapping inside `with_token`, and the ServiceUnavailable non-fatal arm wrapping your `(user_id, refresh)` tuple. If the `Err(_)` arm swallows ServiceUnavailable again, a DB blip during deploy goes back to parking launchers and #1264 silently regresses. A grep for `ServiceUnavailable` in launcher_socket.rs post-rebase is a sufficient check.

**2. Unbounded row minting for never-acking launchers.** An OLD launcher (no TokenRefresh handler) presenting a legacy None-expiry token mints a fresh 30-day row on *every* registration — it never persists, never acks, so the presented token stays None-expiry and re-triggers rotation each reconnect. A flapping old launcher reconnecting every 30s mints ~2.9k rows/day for 30 days each. Suggest a cheap guard in `maybe_issue_launcher_token_refresh`: skip if an unrevoked, unexpired row with the same `(user_id, name)` was created within the last hour (one indexed query). Not a blocker if you'd rather land it as a fast-follow, but it should exist before an old launcher meets a flaky network in prod.

**Accepted/known:** if the ack is lost after the launcher persisted the new token, the old None-expiry row lingers valid indefinitely (never graced). Strictly no worse than pre-PR behavior (those rows were permanent anyway) — fine as a documented follow-up (grace the old row when a registration arrives bearing its replacement, needs an old→new link).

Merge-train proposal given the launcher_socket.rs overlap with my stack: #1266 → #1267 → #1269 (my #1256 stack, all green) → then you rebase #1268 across them and merge. The conflicts land in regions you now own the context for (registration match + select loop).

meawoppl:
Updated for the review note: added a 1h mint-rate guard before issuing a refreshed launcher token. If an unrevoked same-(user,name) token row, excluding the authenticating row, was created within the last hour, registration skips rotation instead of minting another 30-day row. If the cooldown query itself fails, it also skips minting so a DB issue cannot amplify row spam.\n\nReran locally after the change:\n- cargo test -p backend token_refresh -- --nocapture\n- cargo fmt --check\n- mkdir -p frontend/dist && RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets

meawoppl:
Cross-review final read (claude): LGTM at f6bd37a. Verified on the rebased branch itself (not just the diff): the #1264 taxonomy survived (NotFound→Unauthorized / other→ServiceUnavailable in verify_and_get_user_with_token, ServiceUnavailable→fatal:false ack arm in launcher_socket), the #1256 stack pieces are intact (cancel token select arm, touch_launcher stamp, gen-guarded unregister), and the mint cooldown guard is in with the conservative skip-on-DB-error choice — right call: worst case is a delayed rotation, never a mint storm. Merge on green.

PR #1269, opened 2026-07-10
Evict half-open connections via server-side liveness sweep

### LINKED ISSUE #1256: Reconnect race: stale disconnect-cleanup unregisters live launcher/proxy connections, wedging outbound delivery until manual restart

meawoppl:
Part 2 of 2 for #1256 (stacked on #1267). Closes #1256 when merged.

Disconnect detection was purely transport-driven: nothing server-side ever noticed a half-open connection (peer gone, FIN never delivered). The 2026-07-08 incident rode exactly that — a wedged socket survived 9+ hours because no code path ever examined connection age.

**Changes**
- `ProxyConnection` / `LauncherConnection` gain `last_seen` (epoch secs, atomic), stamped at registration and on every decodable inbound frame. Heartbeats already flow on both socket types (proxy 1s, launcher 30s), so a live connection is always fresh.
- New `liveness.rs`: `sweep_stale_connections` evicts + cancels any connection silent past its deadline (proxy 60s, launcher 90s), generation-guarded so a concurrent reconnect's fresh entry is never touched. Eviction reuses the PR-#1267 machinery: entry removed, cancel token fired, socket task exits through normal cleanup, peer reconnects.
- Wired as a `spawn_periodic` background task (30s cadence) alongside the existing sweeps.

Worst-case recovery for the entire #1256 wedge class drops from "manual daemon restart, whenever a human notices" to ≤ ~2 minutes, automatic.

**Tests**: sweep evicts silent proxy+launcher and spares fresh ones (channel kept deliberately open to prove eviction is liveness-driven, not dead-channel-driven); `touch` resets the clock. 144 backend tests green, clippy -Dwarnings, fmt clean.

### DIFF
--- a/backend/src/background.rs
+++ b/backend/src/background.rs
@@ -97,6 +97,28 @@ pub fn spawn_stale_session_cleanup(pool: DbPool, manager: SessionManager) {
     });
 }
 
+/// Evict proxy/launcher connections that have gone silent past their
+/// liveness deadline (see `session_manager/liveness.rs`, #1256). The
+/// eviction cancels each stale connection's socket task, so the client's
+/// reconnect logic recovers automatically.
+pub async fn run_liveness_sweep(app_state: Arc<AppState>) {
+    use crate::handlers::websocket::{
+        LAUNCHER_LIVENESS_DEADLINE_SECS, PROXY_LIVENESS_DEADLINE_SECS,
+    };
+
+    let (proxies, launchers) = app_state.session_manager.sweep_stale_connections(
+        PROXY_LIVENESS_DEADLINE_SECS,
+        LAUNCHER_LIVENESS_DEADLINE_SECS,
+    );
+    if proxies > 0 || launchers > 0 {
+        tracing::warn!(
+            "Liveness sweep evicted {} proxy and {} launcher connection(s)",
+            proxies,
+            launchers
+        );
+    }
+}
+
 /// Query user spend from DB and broadcast to all connected web clients
 pub async fn broadcast_user_spend_updates(app_state: Arc<AppState>) {
     use diesel::prelude::*;
--- a/backend/src/handlers/launchers.rs
+++ b/backend/src/handlers/launchers.rs
@@ -388,6 +388,7 @@ mod tests {
             version: "test".to_string(),
             cancel: tokio_util::sync::CancellationToken::new(),
             gen: 0,
+            last_seen: std::sync::atomic::AtomicU64::new(0),
         }
     }
 
--- a/backend/src/handlers/websocket/launcher_socket.rs
+++ b/backend/src/handlers/websocket/launcher_socket.rs
@@ -182,6 +182,7 @@ pub async fn handle_launcher_socket(socket: WebSocket, app_state: Arc<AppState>)
             version: version.unwrap_or_default(),
             cancel: cancel.clone(),
             gen: 0, // stamped by try_register_launcher
+            last_seen: std::sync::atomic::AtomicU64::new(0), // stamped by try_register_launcher
         },
     );
 
@@ -247,6 +248,9 @@ pub async fn handle_launcher_socket(socket: WebSocket, app_state: Arc<AppState>)
             result = ws_receiver.recv() => {
                 match result {
                     Some(Ok(msg)) => {
+                        // Liveness stamp: any decodable inbound frame proves
+                        // the transport is alive (see liveness.rs).
+                        app_state.session_manager.touch_launcher(&launcher_id);
                         handle_launcher_message(
                             msg,
                             launcher_id,
--- a/backend/src/handlers/websocket/mod.rs
+++ b/backend/src/handlers/websocket/mod.rs
@@ -13,7 +13,8 @@ mod web_client_socket;
 
 pub use session_manager::{
     ForwardHealth, LauncherConnection, ProxySender, SessionId, SessionManager, TunnelError,
-    TunnelIn, WebClientSender,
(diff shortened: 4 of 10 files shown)


meawoppl:
Reviewed the #1269 liveness-sweeper delta. Conditional LGTM once #1267's direct-send bypass blocker is fixed.\n\nChecked the requested seams:\n- sweep vs reconnect: collection is separate from eviction, then each candidate re-reads the current entry and compares gen before dropping the DashMap ref and evicting/canceling. That is the right race shape; a successor entry should be spared.\n- touch placement: stamping after a decodable inbound frame and only after session_key/launcher_id registration is right. Registration initializes last_seen, so the first Register frame does not need a pre-registration touch.\n- deadlines: proxy 60s with 1s heartbeat and launcher 90s with 30s heartbeat are reasonable; 30s sweep cadence gives acceptable worst-case recovery for the half-open class.\n\nOne small fix/nit: sweep_stale_connections returns stale_sessions.len()/stale_launchers.len(), i.e. collected candidates, not actual evictions after the generation re-check. In the exact reconnect race you guarded, the successor is spared but the background log can still say it was evicted. I’d count actual successful evictions instead (or at least rename the log), since this is the path we'll use for incident forensics. Not a behavior blocker if you want to keep it, but it is easy to make precise.

meawoppl:
Delta review update: the actual-eviction count fix is present here (evict_dead_connection / unregister_launcher return bool, and the sweeper counts/logs only successful removals). That part LGTM.\n\nHowever, I fetched pull/1269/head and it still contains the #1267 direct-send bypasses (client_fanout broadcast_shutdown, launcher_registry stop/pause, launchers update/restart). Since #1269 is stacked on #1267, I can't give final LGTM on the train until #1267's branch actually contains the claimed helper-routing fixes.

meawoppl:
LGTM on the rebased liveness branch. I re-fetched meawoppl/liveness-sweeper at 2366720. The sweeper collects candidates, re-checks generation before eviction, drops the map guard before evicting/cancelling, and now counts/logs only real removals via bool returns from evict_dead_connection / unregister_launcher. Touch stamping and the 60s proxy / 90s launcher deadlines still look sane.\n\nThis is good to merge after #1272 lands.

Issue #939, opened 2026-06-04
Harden UI-to-proxy message delivery semantics


meawoppl:
## Summary

The current UI -> backend -> proxy -> agent -> backend -> UI message path is stronger than a plain WebSocket relay: it uses typed protocol enums, ordered frontend sending, backend access checks, a DB-backed `pending_inputs` queue, proxy output sequencing, backend output acknowledgements, DB-backed history, and server timestamps for replay watermarks.

However, the delivery contract is uneven. The system mostly guarantees that input is durable until the proxy receives it, but not necessarily until the underlying agent process accepts it. The browser also has no first-class acknowledgement for submitted input, so the UI relies on optimistic local echoes and later content-based reconciliation.

This issue tracks tightening that path into explicit, end-to-end delivery semantics.

## Current message path

Plain text input currently flows roughly as:

1. `InputBar` reads/clears textarea and emits text + send mode.
2. `SessionView` creates an optimistic pending user message and sends `ClientToServer::ClaudeInput`.
3. Frontend WebSocket code serializes outbound frames through a single queue.
4. Backend validates session mutation permission.
5. Backend increments `sessions.input_seq`, inserts `pending_inputs`, and sends `ServerToProxy::SequencedInput`.
6. Proxy receives `SequencedInput`, pushes it into an internal event channel, and immediately sends `ProxyToServer::InputAck`.
7. Proxy main loop later writes the user message JSONL to Claude/Codex stdin.
8. Agent output is sequenced by the proxy as `ProxyToServer::SequencedOutput`.
9. Backend stores output in `messages`, sends `ServerToProxy::OutputAck`, and broadcasts `ServerToClient::ClaudeOutput` to browsers.
10. Frontend reconciles pending optimistic sends by matching later user/assistant/result output.

Useful refs observed during review:

- `frontend/src/pages/dashboard/session_view/input_bar.rs` - text and upload send initiation
- `frontend/src/pages/dashboard/session_view/component.rs` - optimistic pending user messages
- `frontend/src/pages/dashboard/session_view/websocket.rs` - queued browser WebSocket sender
- `backend/src/handlers/websocket/web_client_socket.rs` - input validation, sequencing, `pending_inputs` insert
- `backend/src/handlers/websocket/message_handlers.rs` - pending input replay, output persistence, output ack
- `backend/src/handlers/websocket/proxy_socket.rs` - `InputAck` deletes pending input rows
[...]

PR #1271, opened 2026-07-10
Backend input idempotency by client_msg_id, outbox resends all unacked

### LINKED ISSUE #1236: Backend idempotency by client_msg_id → close the web-client in-flight-loss window (follow-up to #1235)
## Context
#1235 added a client-side send outbox that makes web-client input delivery reliable across **reconnects**, dup-free: it resends only frames never handed to the transport, so a message the backend already received is never re-sent.

## Remaining gap (deliberately out of #1235's scope)
Two related cases are still uncovered, both requiring server-side idempotency:

1. **In-flight loss:** a frame handed to a socket (or, more precisely, *accepted by the frontend send mpsc queue* — `send_message → true` marks it `transmitted`) that then dies before the WS task writes it to the wire. The outbox marks it `transmitted` and deliberately does **not** resend it (resending would risk a duplicate, since it *might* have reached the backend). So it can be silently lost.

2. To make the outbox able to resend `transmitted`-but-unacked frames on reconnect **without** duplicating, the backend must be idempotent.

## Proposed fix
Dedupe + idempotent re-ack keyed on `client_msg_id`, backend-side:
- Keep a **bounded, per-session** recent-`client_msg_id` map (LRU or short TTL — resends happen within seconds; ~128–256 ids/session is plenty) recording each id's last delivery state (in-flight / `AgentAccepted` / `Failed`).
- On an incoming `AgentInput` with a known `client_msg_id`:
  - terminal `AgentAccepted` → **don't** re-enqueue to the agent; re-emit `InputProgress{AgentAccepted}` to the client so its outbox clears (idempotent ack).
  - terminal `Failed` → re-emit `Failed`.
  - still in-flight → drop the duplicate (original is pending; its ack will come).
- New id → record in-flight, enqueue as today; mark terminal when the proxy reports the stage.

With that, the client outbox can safely resend **all** unacked entries on reconnect (not just un-transmitted ones), closing the in-flight-loss window with at-least-once + idempotent delivery.

This is the web-client twin of the reconnect-ack durability gap tracked under #939 (inter-agent path). Entry points: `backend/src/handlers/websocket/web_client_socket.rs` (AgentInput handler), `SessionManager::enqueue_input`, and the proxy `AgentAccepted` report path.

meawoppl:
Fixes #1236; closes the #939 phase-2/3 remainder.

#1235's client outbox made web input reliable across reconnects but deliberately never resent a frame already handed to a transport — resending might duplicate, so a frame that died in flight with its socket was silently lost. This PR adds the server-side idempotency that makes resending everything safe.

**Backend**
- `pending_inputs.client_msg_id` (migration + model): the delivery-tracking id is persisted with the row, and replay now carries it — delivery tracking survives proxy reconnects instead of degrading to content reconciliation, and the idempotency gate keeps recognizing replayed ids.
- New `input_dedup` module: bounded per-session tracker (256 ids) of `client_msg_id → InFlight/Accepted/Failed`. `check_and_record_input` is one entry-lock op, so concurrent duplicates can't both read `New`.
- `AgentInput` gate in the web-client socket: duplicate in-flight → dropped (original's acks will arrive); duplicate terminal → the terminal `InputProgress` stage is re-emitted so the client's outbox resolves (idempotent ack); new → proceeds. DB backstop: after a backend restart the tracker is empty, so a `pending_inputs` row with the same id counts as in-flight.
- The proxy's `InputProgressAck` terminal stages are recorded into the tracker.

**Frontend**
- The reconnect flush resends **every unresolved** outbox entry, including transmitted-but-unacked ones (`untransmitted()` → `unresolved()`). At-least-once + idempotent = the in-flight-loss window is closed and "Sending…" always resolves truthfully.

**Documented residual**: a terminal state recorded only in memory is lost if the backend restarts before the client's ack round-trip completes AND the pending row was already deleted — a resend then re-delivers. Rare (double failure in a seconds-wide window), and a visible duplicate beats a silent loss.

**Tests**: dedup transitions (in-flight → accepted, terminal-for-untracked, per-session boundedness, independence), outbox resend-all + never-resend-resolved. Backend 141, frontend 379, clippy `-Dwarnings` both targets, fmt clean.

### DIFF
--- a/backend/migrations/2026-07-10-144459_add_client_msg_id_to_pending_inputs/down.sql
+++ b/backend/migrations/2026-07-10-144459_add_client_msg_id_to_pending_inputs/down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE pending_inputs DROP COLUMN client_msg_id;
--- a/backend/migrations/2026-07-10-144459_add_client_msg_id_to_pending_inputs/up.sql
+++ b/backend/migrations/2026-07-10-144459_add_client_msg_id_to_pending_inputs/up.sql
@@ -1,0 +1,5 @@
+-- Delivery-tracking id for pending inputs (#1236). Carried from the web
+-- client's outbox so (a) replay after a proxy reconnect preserves delivery
+-- tracking instead of degrading to content reconciliation, and (b) the
+-- backend can deduplicate an input the client resends after a reconnect.
+ALTER TABLE pending_inputs ADD COLUMN client_msg_id UUID;
--- a/backend/src/handlers/websocket/launcher_socket.rs
+++ b/backend/src/handlers/websocket/launcher_socket.rs
@@ -549,6 +549,7 @@ fn handle_launcher_message(
                     seq_num: next_seq,
                     content: serde_json::to_string(&content_value).unwrap_or_default(),
                     send_mode: None,
+                    client_msg_id: None,
                 };
                 let _ = diesel::insert_into(pending_inputs::table)
                     .values(&new_input)
--- a/backend/src/handlers/websocket/message_handlers.rs
+++ b/backend/src/handlers/websocket/message_handlers.rs
@@ -99,9 +99,11 @@ pub fn replay_pending_inputs_from_db(
             seq: input.seq_num,
             content,
             send_mode: input.send_mode.as_deref().and_then(parse_send_mode),
-            // Replayed inputs aren't delivery-tracked (the original browser
-            // row is long gone); UI falls back to content reconciliation.
-            client_msg_id: None,
+            // Persisted with the row (#1236) so replay keeps delivery
+            // tracking: the proxy's InputProgressAck still resolves the
+            // browser's outbox entry, and the idempotency gate keeps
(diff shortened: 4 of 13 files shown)


meawoppl:
LGTM from the focused #1236 review.\n\nWhat I checked:\n- Permission/mutator check runs before the dedup record, so a denied input does not poison a client_msg_id for a later authorized retry.\n- check_and_record_input only holds the DashMap entry lock over the in-memory VecDeque check/insert; pending_input_exists does its DB call outside that lock. No re-entrant SessionManager path there.\n- Reconnect resend-all is scoped to WebSocketConnected; normal sends still go through the one-shot transmitted path, so I don't see a resend storm path. Register is enqueued/sent before Connected is emitted, and outbox flush uses the same ordered socket queue, so replayed inputs should arrive after Register on the backend.\n- The residual duplicate window after backend restart + pending row deletion is documented accurately. Dup-beats-loss is the right side of the contract for that double-failure case.\n\nNo blockers found.

PR #1272, opened 2026-07-10
Unify connection identity with generation guards and dead-connection eviction

meawoppl:
Part 1 of 2 for #1256. Replaces #1267, which GitHub auto-closed (its base branch was deleted by the #1266 merge) and froze at a pre-review-fix SHA — this PR contains the same two commits, now including the review fix routing ALL registry sends through the evicting helpers.

The 2026-07-08 incident (#1256): a reconnect flap left connections alive inbound but unroutable outbound for 9+ hours. This PR completes the identity model:

**One registry entry, one identity.** The proxy `sessions` map stores `ProxyConnection { sender, gen, cancel }` — the separate `connection_gen` map is deleted (no two-map disagreement possible; the tunnel opener's dual lookup collapses to one), and `unregister_session` becomes an atomic `remove_if`.

**Launchers get the same guard.** A launcher reconnect reuses its stable `launcher_id`, so the old socket's late cleanup could remove the new socket's registry entry. `LauncherConnection` now carries a generation; cleanup is compare-and-remove.

**Dead connections are evicted and closed, not tolerated.** Every entry carries a `CancellationToken`. A failed send to a registered connection evicts the entry (gen-guarded, sparing successors) and cancels the socket task; the peer's reconnect logic takes over. Per Codex review, every send path now routes through the evicting helpers — `broadcast_shutdown` (keys collected first; eviction never runs under the iterator's shard lock), `stop_session_on_launcher` / `pause_session_on_launcher` (target id resolved via `launcher_running_session`, then evicting send), and the `UpdateAndRestart` handler (no more raw sender clone).

**Tests**: send-failure eviction + cancel, gen-guarded successor sparing, launcher same-id stale-unregister no-op, launcher send-failure eviction releasing the dedup slot; 142 backend tests, clippy `-Dwarnings`, fmt.

### DIFF
--- a/backend/src/handlers/launchers.rs
+++ b/backend/src/handlers/launchers.rs
@@ -381,6 +381,8 @@ mod tests {
             running_sessions: Vec::new(),
             working_directory: None,
             version: "test".to_string(),
+            cancel: tokio_util::sync::CancellationToken::new(),
+            gen: 0,
         }
     }
 
--- a/backend/src/handlers/launchers.rs
+++ b/backend/src/handlers/launchers.rs
@@ -327,7 +327,7 @@ pub async fn probe_agents(
     CurrentUserId(user_id): CurrentUserId,
     Path(launcher_id): Path<Uuid>,
 ) -> Result<Json<ProbeAgentsResponse>, AppError> {
-    let sender = {
+    {
         let launcher = app_state
             .session_manager
             .launchers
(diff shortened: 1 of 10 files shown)


meawoppl:
Still one same-class blocker after the fresh branch review. The original four bypasses are fixed in #1272, but there are still registry sends that bypass the evicting helpers:\n\n1. backend/src/handlers/launchers.rs probe_agents clones launcher.sender and calls sender.send(ServerToLauncher::ProbeAgents { request_id }) directly. If that channel is closed, it cancels the probe request and returns 502, but it leaves the stale launcher registry/dedup entry behind. This should route through send_to_launcher after the authorization check, just like UpdateAndRestart.\n\n2. backend/src/handlers/websocket/session_manager/proxy_lifecycle.rs disconnect_session reads conn.sender and sends SessionTerminated directly. If the channel is closed, it returns false but does not evict/cancel the dead proxy connection. Under the new invariant this should use send_to_connected_session (or a termination-specific helper that evicts on send failure).\n\nThe updated branch does fix broadcast_shutdown, stop/pause, and UpdateAndRestart; those look good. I want these two residual direct registry sends cleaned up before final LGTM, otherwise the invariant remains partial.

meawoppl:
LGTM. I re-fetched #1272 at bd2c5fa and checked the exact changed paths. The original bypasses plus the two residuals are now routed through evicting helpers:\n\n- broadcast_shutdown collects session/launcher ids before sending via send_to_connected_session/send_to_launcher.\n- stop_session_on_launcher and pause_session_on_launcher resolve the launcher id first, then send via send_to_launcher.\n- UpdateAndRestart and ProbeAgents keep the auth check under a scoped get guard, then send via send_to_launcher.\n- disconnect_session uses send_to_connected_session.\n\nThe remaining direct sender uses I see are the helper internals / websocket writer loops / replay relay paths, not registry-bypass call sites. Good to merge once green.

PR #1273, opened 2026-07-10
Transactional file uploads: temp-file commit, result acks, gated prompt

meawoppl:
#939 phase 4. Uploads were fire-and-forward on all three legs with zero acknowledgement: the proxy wrote bytes **directly to the final path** (an agent could read a truncated file mid-upload), every failure was a local log line, and the browser dispatched the prompt referencing the file unconditionally — so a dropped chunk still produced a prompt about a file that doesn't fully exist.

## Changes

**Protocol** — new `FileUploadResultFields { upload_id, success, error }`; carried as `ProxyToServer::FileUploadResult` and `ServerToClient::FileUploadResult` (additive; old peers ignore/skip them).

**Proxy** — writes to a hidden temp path (`.{name}.{id8}.upload`) and renames into place only when all chunks are flushed: the commit barrier. Every failure path (create, decode, write, flush, rename) now cleans up the partial file, drops the state, and reports a failure result; completion reports success. Write errors previously didn't even bump the chunk counter — the upload stalled forever, silently.

**Backend** — relays the proxy's result to the session's web clients, and synthesizes failures it can already detect: invalid chunk count, size-cap rejections (previously *silent drops*), abort outcomes, and proxy-offline. Uploads are now connected-only (`send_to_connected_session`): the old queue-through path capped at 100 in-memory frames, so \"queueing\" a multi-thousand-chunk upload for a disconnected proxy silently dropped most of it — failing fast is honest.

**Frontend** — the InputBar hands the composed prompt + upload ids to the SessionView instead of dispatching it; the view holds the prompt until every id commits, then dispatches through the normal outbox path. Failure surfaces as a transcript error (\"your message was not sent\") — the agent is never told about a file that isn't fully on disk. Results that race ahead of the prompt handoff are stashed and consumed. A 45s timeout falls back to pre-transactional behavior for old proxies that never ack.

## Compat matrix
- new browser + old proxy: timeout fallback → today's behavior
- old browser + new proxy/backend: extra result frames are unknown-variant no-ops
- new browser + new stack: full transaction

Tests: existing upload validation suite extended for the new signatures (5 e2e call sites); full workspace clippy `-Dwarnings`, wasm clippy, backend 148 / frontend 379 tests, fmt.

### DIFF
--- a/backend/src/handlers/websocket/proxy_socket.rs
+++ b/backend/src/handlers/websocket/proxy_socket.rs
@@ -403,6 +403,17 @@ fn handle_proxy_message(
                 );
             }
         }
+        ProxyToServer::FileUploadResult(result) => {
+            // Relay the upload's terminal outcome to the session's web
+            // clients — the uploader is holding back the prompt that
+            // references the file until this arrives (#939 phase 4).
+            if let Some(key) = session_key {
+                session_manager.broadcast_to_web_clients(
+                    key,
+                    shared::ServerToClient::FileUploadResult(result),
+                );
+            }
+        }
         ProxyToServer::SessionStatus { .. } => {}
         ProxyToServer::ForwardStatus(status) => {
             if let Some(session_id) = *db_session_id {
--- a/backend/src/handlers/websocket/uploads.rs
+++ b/backend/src/handlers/websocket/uploads.rs
@@ -31,10 +31,23 @@ pub(super) struct PendingUpload {
     received_bytes: u64,
 }
 
+/// Tell the uploading web client its upload definitively failed (#939
+/// phase 4) — it is holding back the prompt that references the file.
+fn send_upload_failure(tx: &super::WebClientSender, upload_id: String, error: &str) {
+    let _ = tx.send(shared::ServerToClient::FileUploadResult(
+        shared::FileUploadResultFields {
+            upload_id,
+            success: false,
+            error: Some(error.to_string()),
+        },
+    ));
+}
+
 #[allow(clippy::too_many_arguments)]
 pub(super) fn handle_file_upload_start(
     session_manager: &SessionManager,
     session_key: &Option<SessionId>,
+    tx: &super::WebClientSender,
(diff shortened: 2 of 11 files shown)


meawoppl:
Blocker: duplicate FileUploadStart with the same upload_id leaks the previous temp file on the proxy.\n\nIn claude-session-lib/src/proxy_session/mod.rs, the Start arm creates a new .*.upload temp file and then does active_uploads.insert(upload_id, FileReceiveState { ... temp_path, ... }) around lines 1119-1134. If a client sends another FileUploadStart with the same upload_id before the first completes, HashMap::insert drops the old FileReceiveState and file handle but never removes old_state.temp_path. That leaves a hidden .upload file in the working directory, violating the transactional cleanup/no-temp-leak invariant.\n\nThis is client-controlled wire input, so I would not rely on the Yew client generating fresh UUIDs. Please reject duplicate starts or explicitly abort/remove the existing upload before returning. I think the cleanest behavior is: if active_uploads.contains_key(&upload_id), call fail_upload(state, upload_id.clone(), "duplicate upload start") and return without creating a new temp file. A backend-side duplicate-start guard would also be fine, but the proxy still needs to be robust because it owns the temp file.

meawoppl:
LGTM after the duplicate-start fix. I re-fetched at 913b1ca and checked the upload path again.\n\nWhat I verified:\n- Duplicate FileUploadStart now removes the old active_uploads entry, drops the handle, and deletes the old temp file before creating the replacement, so the proxy no longer leaks .upload files.\n- fail_upload removes state before deleting temp/reporting failure, so the borrow discipline is clean.\n- Backend connected-only upload forwarding is the right behavior: queuing a multi-frame upload into the bounded pending queue was an illusion, and failures now get a typed FileUploadResult back to the uploader.\n- Frontend only dispatches the composed prompt after all referenced uploads succeed, handles early results, surfaces terminal failures without sending the prompt, and the timeout fallback preserves compatibility with old proxies.\n- New wire variants are additive.\n\nNo blockers found.

PR #1274, opened 2026-07-10
Bound per-connection outbound channels with evict-on-overflow

### LINKED ISSUE #939: Harden UI-to-proxy message delivery semantics

meawoppl:
#939 phase 5, the final PR of the connection-lifecycle series. Closes #939.

Every per-connection outbound channel (backend → proxy / launcher / web client) was an unbounded mpsc: a slow-but-alive consumer let the queue grow without limit — the last unbounded memory growth path on the connection plane.

## Design

New `ConnSender<T>`: a bounded channel whose `send` is a non-blocking `try_send`, with **overflow deliberately indistinguishable from a closed channel**. That means no new error handling anywhere: the existing #1256 machinery (send-failure → gen-guarded evict + cancel token → socket closes → peer reconnects) handles an over-budget consumer exactly like a wedged one — operationally the same thing.

## Capacities

- **Proxy: 16 384 frames** — sized so a full maximum-size upload (10 MiB at ~1 KiB decoded/chunk ≈ 10 240 frames) fits even when the proxy's downlink is much slower than the uploader's; the chunk fan-in is the largest legitimate burst on this channel. Tunnel streams are self-limited by their flow-control window.
- **Launcher: 1 024** — reconcile/launch/schedule traffic is a few frames per 30s heartbeat.
- **Web client: 4 096** — output is agent-paced; history replay is one batched frame.

Worst-case memory is now bounded per connection instead of unbounded, and the failure mode is a clean reconnect instead of OOM.

Tunnel inbox channels are explicitly NOT converted (comment added) — they're flow-controlled by the tunnel window, not connection senders.

**Tests**: all 153 backend tests green (the pending-queue overflow test's channel is sized above the replay burst it deliberately provokes, with a comment); workspace clippy `-Dwarnings`, fmt.

### DIFF
--- a/backend/src/handlers/launchers.rs
+++ b/backend/src/handlers/launchers.rs
@@ -374,10 +374,9 @@ pub async fn probe_agents(
 mod tests {
     use super::*;
     use crate::handlers::websocket::LauncherConnection;
-    use tokio::sync::mpsc;
 
     fn launcher_for(user_id: Uuid, hostname: &str) -> LauncherConnection {
-        let (sender, _rx) = mpsc::unbounded_channel();
+        let (sender, _rx) = crate::handlers::websocket::conn_channel(64);
         LauncherConnection {
             sender,
             launcher_name: format!("launcher-{}", hostname),
--- a/backend/src/handlers/websocket/launcher_socket.rs
+++ b/backend/src/handlers/websocket/launcher_socket.rs
@@ -5,7 +5,6 @@ use shared::{
     ServerToProxy, SessionStatus,
 };
 use std::sync::Arc;
-use tokio::sync::mpsc;
 use tracing::{debug, error, info, warn};
 use uuid::Uuid;
 
@@ -318,7 +317,8 @@ pub async fn handle_launcher_socket(socket: WebSocket, app_state: Arc<AppState>)
     }
 
     // Create channel for sending messages to this launcher
-    let (tx, mut rx) = mpsc::unbounded_channel::<ServerToLauncher>();
+    let (tx, mut rx) =
+        super::session_manager::conn_channel::<ServerToLauncher>(super::LAUNCHER_CHANNEL_CAPACITY);
     let tx_for_sync = tx.clone();
 
     // Server-side kill switch: fired by the SessionManager if it evicts this
--- a/backend/src/handlers/websocket/mod.rs
+++ b/backend/src/handlers/websocket/mod.rs
@@ -12,9 +12,10 @@ mod uploads;
 mod web_client_socket;
 
 pub use session_manager::{
-    ForwardHealth, LauncherConnection, ProxySender, SessionId, SessionManager, TunnelError,
-    TunnelIn, WebClientSender, LAUNCHER_LIVENESS_DEADLINE_SECS, LIVENESS_SWEEP_INTERVAL_SECS,
-    PROXY_LIVENESS_DEADLINE_SECS,
+    conn_channel, ConnSender, ForwardHealth, LauncherConnection, ProxySender, SessionId,
+    SessionManager, TunnelError, TunnelIn, WebClientSender, LAUNCHER_CHANNEL_CAPACITY,
+    LAUNCHER_LIVENESS_DEADLINE_SECS, LIVENESS_SWEEP_INTERVAL_SECS, PROXY_CHANNEL_CAPACITY,
+    PROXY_LIVENESS_DEADLINE_SECS, WEB_CLIENT_CHANNEL_CAPACITY,
 };
 
 use axum::{
--- a/backend/src/handlers/websocket/proxy_socket.rs
+++ b/backend/src/handlers/websocket/proxy_socket.rs
@@ -12,7 +12,6 @@ use diesel::prelude::*;
 use shared::{ProxyToServer, ServerToClient, ServerToProxy, SessionEndpoint, SessionStatus};
 use std::ops::ControlFlow;
 use std::sync::Arc;
-use tokio::sync::mpsc;
 use tracing::{error, info, warn};
 use uuid::Uuid;
 
@@ -21,7 +20,8 @@ pub async fn handle_session_socket(socket: WebSocket, app_state: Arc<AppState>) 
     let db_pool = app_state.db_pool.clone();
     let conn = ws_bridge::server::into_connection::<SessionEndpoint>(socket);
     let (mut ws_sender, mut ws_receiver) = conn.split();
-    let (tx, mut rx) = mpsc::unbounded_channel::<ServerToProxy>();
+    let (tx, mut rx) =
+        super::session_manager::conn_channel::<ServerToProxy>(super::PROXY_CHANNEL_CAPACITY);
 
     let mut session_key: Option<SessionId> = None;
     let mut db_session_id: Option<Uuid> = None;
--- a/backend/src/handlers/websocket/replay.rs
+++ b/backend/src/handlers/websocket/replay.rs
@@ -291,9 +291,9 @@ mod replay_tests {
     /// injection without standing up a full WS round-trip.
     fn capture_sender() -> (
         super::WebClientSender,
-        tokio::sync::mpsc::UnboundedReceiver<shared::ServerToClient>,
+        tokio::sync::mpsc::Receiver<shared::ServerToClient>,
     ) {
-        tokio::sync::mpsc::unbounded_channel::<shared::ServerToClient>()
+        crate::handlers::websocket::conn_channel::<shared::ServerToClient>(64)
     }
 
     /// The fix's whole point: `replay_history` with a `replay_after`
--- a/backend/src/handlers/websocket/session_manager/client_fanout.rs
+++ b/backend/src/handlers/websocket/session_manager/client_fanout.rs
@@ -98,13 +98,12 @@ impl SessionManager {
 mod tests {
     use super::super::test_support::make_client_msg;
     use super::*;
-    use tokio::sync::mpsc;
 
     #[test]
     fn broadcast_to_web_clients() {
         let mgr = SessionManager::new();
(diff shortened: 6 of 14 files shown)


meawoppl:
LGTM. I reviewed the bounded-channel conversion and don't see a blocker.\n\nWhat I checked:\n- Overflow mapping to SendError is consistent with the #1256 evict/cancel machinery. For the connection registries, treating a full outbound queue as wedged is the right policy; drop-frame would make delivery state ambiguous.\n- Capacity sizing looks reasonable. Proxy 16k is safely above the legitimate upload burst for the default 10MiB / 1KiB chunk path, and the pending replay queue is only 100 frames, so replay cannot overflow the proxy channel during register_session. Launcher 1024 and web client 4096 are comfortably above normal use.\n- The connection registries are converted: proxy, launcher, web-client/user-client senders now use ConnSender. The remaining unbounded channels are internal agent/proxy/io channels or tunnel inboxes. I agree tunnel inboxes are out of scope because they are governed by the tunnel window/stream lifecycle.\n- The recv loops remain await-based on bounded Receiver, and call sites intentionally keep nonblocking try_send semantics via ConnSender::send.\n\nNo blockers found. Merge on green.

meawoppl:
LGTM. I checked the overflow-as-closed path against the #1256 eviction/cancel machinery, the capacity sizing, and the remaining unbounded channels. Proxy capacity is comfortably above pending replay and the expected upload burst, launcher/web client bounds are reasonable, tunnel inboxes remain excluded because their flow-control window is the actual bound, and the send-task recv loops keep the same await-based behavior. No blockers; merge on green.

PR #1276, opened 2026-07-11
Drop empty loop-priming user echoes from transcript

### LINKED ISSUE #715: Empty / loop-priming Claude user messages should not render in the transcript
## Observed

Claude's stdout occasionally emits `ClaudeOutput::User` echoes whose text content is **empty** (or near-empty / whitespace-only). These appear to be priming events the CLI injects to bootstrap the message loop, not actual user-typed input.

The frontend currently renders them as blank "user" bubbles in the transcript, which:

- Visually fragments the conversation (a stray empty message between agent turns).
- Confuses the eye when scanning history — it looks like the user hit Send without typing anything.
- Bloats the user-message count for retention math.

## Expected

Drop these echoes in the proxy's output forwarder before they reach the frontend, the same way we already drop `<system-reminder>` echoes (2.5.23) — i.e. detect at the [proxy_session/wiggum.rs:255](https://github.com/meawoppl/agent-portal/blob/main/claude-session-lib/src/proxy_session/wiggum.rs#L255) layer (`is_system_reminder_user_echo`) and add a sibling predicate for empty / whitespace-only user echoes.

## Proposed predicate

Drop a `ClaudeOutput::User` when **all** of its text content blocks (a) trim to empty, or (b) contain only the `<system-reminder>` wrapper text. Tool-result blocks and non-text blocks should not be considered — those are real tool-result echoes and have to render.

Pseudocode:

```rust
fn is_empty_user_echo(output: &ClaudeOutput) -> bool {
    let ClaudeOutput::User(msg) = output else { return false };
    let texts: Vec<&str> = msg.message.content.iter()
        .filter_map(|b| match b {
            ContentBlock::Text(t) => Some(t.text.as_str()),
            _ => None,
        })
        .collect();
    if texts.is_empty() {
        // no text blocks at all (just tool_results, etc.) → not what we're filtering
        return false;
    }
    texts.iter().all(|t| t.trim().is_empty())
}
```

Wire it into the same drop arm in `wiggum.rs` next to the existing `<system-reminder>` filter so the two cases share the "echo Claude emitted, not real user input" classification.

## Related history

- 2.5.23 — added the `<system-reminder>` user-echo filter (different priming text, same problem).
- 2.5.28 — fixed Codex's parallel "synthetic user echo" issue via shape-based dispatch (PR #691, #702).

This issue is the third instance of the same shape: an agent CLI emits an echo of its own loop-priming input as a `User` message, and the frontend has no way to tell it apart from a real user prompt.

meawoppl:
Fixes #715.

The Claude CLI occasionally emits `User` echoes whose text content is empty or whitespace-only — loop-priming events, not real input. They rendered as blank user bubbles that fragment the transcript and inflate the user-message count.

New `is_empty_user_echo` predicate in the proxy's output path, wired next to the existing `<system-reminder>` echo drop (same classification: an echo Claude emitted, not something the user typed). Per the issue spec, it matches only when the echo has at least one text block and every text block trims to empty — echoes with tool-result or other non-text blocks are real content and pass through untouched.

Gates: clippy -Dwarnings, claude-session-lib tests, fmt.

### DIFF
--- a/claude-session-lib/src/proxy_session/wiggum.rs
+++ b/claude-session-lib/src/proxy_session/wiggum.rs
@@ -530,6 +530,60 @@ mod tests {
         }
     }
 
+    /// Build a `ClaudeOutput::User` frame from raw content blocks, the same
+    /// way the CLI ships them.
+    fn user_frame(content: serde_json::Value) -> ClaudeOutput {
+        serde_json::from_value(serde_json::json!({
+            "type": "user",
+            "session_id": "00000000-0000-0000-0000-000000000000",
+            "message": { "role": "user", "content": content }
+        }))
+        .expect("parses as ClaudeOutput")
+    }
+
+    /// #715 predicate contract: drop ONLY user echoes whose every text
+    /// block is empty/whitespace — anything carrying real content (tool
+    /// results, images, non-empty text, or no text blocks at all) must
+    /// render.
+    #[test]
+    fn empty_user_echo_predicate_contract() {
+        // Whitespace-only text → dropped.
+        assert!(is_empty_user_echo(&user_frame(serde_json::json!([
+            { "type": "text", "text": "   \n\t" }
+        ]))));
+        assert!(is_empty_user_echo(&user_frame(serde_json::json!([
+            { "type": "text", "text": "" },
+            { "type": "text", "text": " " }
+        ]))));
+
+        // No content blocks at all → NOT what we filter.
+        assert!(!is_empty_user_echo(&user_frame(serde_json::json!([]))));
+
+        // Tool-result / non-text block → real content, passes through.
+        assert!(!is_empty_user_echo(&user_frame(serde_json::json!([
+            { "type": "tool_result", "tool_use_id": "t1", "content": "ok" }
+        ]))));
+
+        // Mixed empty text + non-text → passes through.
+        assert!(!is_empty_user_echo(&user_frame(serde_json::json!([
+            { "type": "text", "text": "" },
+            { "type": "tool_result", "tool_use_id": "t1", "content": "ok" }
+        ]))));
+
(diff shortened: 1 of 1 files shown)


meawoppl:
Cross-review (codex): code shape looks right: the empty-echo filter is after the system-reminder drop, and the  guard correctly keeps no-text user frames (tool results / other structured echoes) flowing through. One requested fix before merge: add unit coverage for  itself. I ran 
running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 36 filtered out; finished in 0.00s and it executed 0 tests, so the new behavior is currently unpinned. Minimal cases I’d want: whitespace-only text user => true; no content blocks => false; tool_result/non-text block => false; mixed empty text + non-text => false; non-empty text => false. Also note this branch predates the SDK bump (#1278), so rebase onto current main before merging and make sure the wiggum  fixture picks up the new 2.1.160 fields.

meawoppl:
Correction to my previous comment: shell backticks mangled the identifiers. The requested unit coverage is for `is_empty_user_echo`:

- whitespace-only text user => true
- no content blocks => false
- tool_result/non-text block => false
- mixed empty text + non-text => false
- non-empty text => false

The stale-branch note is also specifically about the `ResultMessage` test fixture after the 2.1.160 SDK bump.

meawoppl:
Rebased over post-#1278 main (ResultMessage fixture now carries the 2.1.160 fields via the rebase) and added the requested predicate coverage as `empty_user_echo_predicate_contract`: whitespace-only → true; no blocks → false; tool-result block → false; mixed empty-text + non-text → false; non-empty text → false; plus a non-user-frame guard. All green locally (clippy -Dwarnings, tests, fmt).

Issue #1050, opened 2026-06-16
Surface MCP elicitation server name once codex-codes exposes it


meawoppl:
## Location
`codex-session-lib/src/handler.rs:319-332` (`ServerRequest::McpServerElicitationRequest`)

## Context
The handler hard-defaults `server_name` to `"(unknown)"` because the typed `McpServerElicitationRequestParams` (Form/Url variants) does not expose a server identifier. The rendered permission text is therefore always `MCP server `(unknown)` is asking …`.

```rust
// TODO(SDK): if upstream adds a server identifier to the typed params,
// surface it here and widen `CodexPermissionInput::McpElicitation` to match.
let input = CodexPermissionInput::McpElicitation {
    server_name: "(unknown)".to_string(),
};
```

## Work
1. File/track upstream (rust-code-agent-sdks) to add a server-name field to the elicitation params.
2. Once available, read it here and widen `CodexPermissionInput::McpElicitation` so the permission card names the real server.

meawoppl:
Checked while bumping codex-codes for #827: codex-codes 0.143.0 still does not expose a server identifier on the exact `McpServerElicitationRequestParams` request type. It does add/keep `server_name` on related remote-control status notifications, but that does not let the permission classifier populate the MCP elicitation dialog. Leaving this open; #827 will keep the current `(unknown)` fallback.

meawoppl:
Triage note: rechecked during the codex-codes 0.143.4 SDK cleanup. MCP elicitation params still do not expose a server/origin field in the upstream Codex schema; the matching SDK issue was closed upstream-blocked. Keeping this open as the focused portal tracker for surfacing the real server name once upstream Codex exposes it.

meawoppl:
Re-audited after rust-code-agent-sdks#302. The resnapshot is now merged and byte-identical to `openai/codex@main`, but the published app-server JSON schema still models `McpServerElicitationRequestParams` only as the mode-tagged form/openai-form/url payload and omits the Rust source struct's `serverName` / `threadId`. Consequently generated `codex-codes` cannot expose a typed server name yet without inventing fields outside its schema source of truth. This remains upstream-blocked on OpenAI's schema export including those fields; #302 does not unblock it.

PR #1277, opened 2026-07-11
Fix Codex file change permissions

### LINKED ISSUE #827: Codex FileChange: in-progress event drops on parse; permission dialog shows item id instead of filenames
### 1. Permission dialog shows the raw item id, not the affected files

When Codex requests permission to apply a file change, the dialog looks like:

> ⚠️ **Permission Required**
> Tool: FileChange
> File change for item `call_apLovlbfsFz11MCYpiVcv0UK`
> (diff streamed above under this item id)
> [Allow] [Deny]

The user has no idea which file is about to be edited until they click Allow.

### 2. The `item.started` event referenced above never actually renders

The "diff streamed above" hint is wrong: the matching `item.started` carrying the file change never renders a card. Here is the raw frame the proxy sent for a recent prompt:

```json
{
  "_created_at": "2026-05-18T23:04:21.140Z",
  "item": {
    "changes": [
      {
        "diff": "@@ -136,2 +136,3 @@\n     let hostname = props.session.hostname.clone();\n+    let session_agent_type = props.session.agent_type;\n …",
        "kind": {"type": "update"},
        "path": "/home/meawoppl/repos/agent-portal-2/frontend/src/components/schedule_dialog.rs"
      }
    ],
    "id": "call_apLovlbfsFz11MCYpiVcv0UK",
    "status": "inProgress",
    "type": "fileChange"
  },
  "type": "item.started"
}
```

Nothing appears in the chat for this frame.

## Root cause (single)

Both symptoms are the same bug in `frontend/src/components/codex_renderer.rs:243-247`:

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct FileChange {
    pub path: Option<String>,
    pub kind: Option<String>,
}
```

Two mismatches with the actual Codex wire shape:

1. **`kind` is `Option<String>` but the wire ships `{"type": "update"}`** (an object). Serde fails to deserialize the variant; the whole `CodexEvent::ItemStarted { item: CodexItem::FileChange { … } }` parse falls back to whichever next arm matches (likely none → silent drop in `codex_message_renderer` at `codex_renderer.rs:264-280`).
2. **`diff` is not a field at all.** Even if `kind` were tolerant, `render_file_change` (`codex_renderer.rs:498-539`) wouldn't have a diff to render — it currently only emits the path + kind chip per change, no diff body. So the rich content the user is about to approve is just discarded.

Because the rendered card never appears, the `PermissionHandler` falls back to its no-files-known text ("File change for item `<call_id>` (diff streamed above under this item id)"), which sends the user into the Allow/Deny prompt blind.

## Fix shape

In `frontend/src/components/codex_renderer.rs`:

1. Change `FileChange.kind` to a typed enum (or `Option<KindWrapper>` with the existing object shape) — the SDK at `codex-codes` may already expose this; if not, file an upstream issue per `CLAUDE.md`'s \"Prefer typed interfaces; push schema gaps upstream\" rule, and use a small local `#[serde(rename_all = \"snake_case\")] enum FileChangeKind { Add, Update, Delete }` plus `#[serde(deny_unknown_fields = false)] struct KindWrapper { type: FileChangeKind }` as the hotfix.
2. Add `pub diff: Option<String>` to the `FileChange` struct, and feed it to a `<DiffCard source=DiffSource::Unified { text: diff } file_path=path kind=kind />` in `render_file_change` (the unified `DiffCard` was just landed by #823/#825 specifically so Codex file-change rendering can stop being naked text — use it here too).

In `frontend/src/pages/dashboard/session_view/permission_handler.rs`:

3. When the pending permission references a `FileChange` and matching items are in the session history, harvest the `path` list and show it in the dialog headline: \"Allow Codex to edit **schedule_dialog.rs**?\" (or list multiple paths if several). Falling back to the call id is fine when no matching item is in scope, but in the common case the matching `item.started` has already streamed.

## Regression coverage

- Existing tests in `codex_renderer::tests` (`item_file_change_snake_case`, `item_file_change_camel_case`) use the string-form `\"kind\":\"update\"` — extend them to also pin the object form `\"kind\":{\"type\":\"update\"}` so this can't silently regress when the SDK schema rotates again.
- Add a test that asserts the rendered HTML for an `item.started` with a `fileChange` containing `diff` includes a `.diff-card` element (proves the DiffCard wiring lands).

meawoppl:
## Summary
- keep the Codex FileChange object-kind/diff typed path on top of the workspace SDK bump from #1278
- enrich Codex FileChange permission requests with filenames resolved from matching item events
- box the app-server CodexItem wrapper to keep clippy clean without a large_enum_variant allow

## Tests
- cargo test -p shared codex_permission_input_file_change --lib
- cargo test -p frontend enrich_codex_file_change_permission_resolves_paths_from_item_events --lib
- cargo test -p frontend format_codex_file_change_permission_prefers_paths --lib
- cargo test -p frontend event_item_started_with_file_change_no_longer_silently_drops --lib
- cargo test -p codex-session-lib --lib
- RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets
- cargo check -p shared --locked

Fixes #827

Notes: checked codex-codes 0.143.x against #1050/#1146. MCP elicitation request params still do not expose server_name; app-server ContextCompaction/CollabAgentToolCall typed variants are present and already used by frontend main.

### DIFF
--- a/codex-session-lib/src/classifier.rs
+++ b/codex-session-lib/src/classifier.rs
@@ -153,6 +153,7 @@ fn classify_request(request_id: String, request: ServerRequest) -> Vec<AgentOutp
             request_id,
             CodexPermissionInput::FileChange {
                 item_id: p.item_id,
+                paths: vec![],
                 reason: p.reason,
                 grant_root: p.grant_root,
             },
--- a/frontend/src/components/codex_renderer.rs
+++ b/frontend/src/components/codex_renderer.rs
@@ -138,27 +138,29 @@ fn render_item(item: Option<&CodexItem>, completed: bool, session_id: Uuid) -> H
             // here to avoid duplication.
             ThreadItem::UserMessage(_) => html! {},
         },
-        CodexItem::AppServer(AppServerThreadItem::ContextCompaction { .. }) => {
-            render_context_compaction_item(completed)
-        }
-        CodexItem::AppServer(AppServerThreadItem::CollabAgentToolCall {
-            agents_states,
-            model,
-            prompt,
-            reasoning_effort,
-            status,
-            tool,
-            ..
-        }) => render_collab_agent_tool_call(
-            tool,
-            model.as_deref(),
-            reasoning_effort.as_ref(),
-            status,
-            prompt.as_deref(),
-            agents_states,
-            completed,
-        ),
-        CodexItem::AppServer(_) => html! {},
+        CodexItem::AppServer(item) => match item.as_ref() {
+            AppServerThreadItem::ContextCompaction { .. } => {
+                render_context_compaction_item(completed)
+            }
+            AppServerThreadItem::CollabAgentToolCall {
+                agents_states,
+                model,
+                prompt,
+                reasoning_effort,
+                status,
+                tool,
+                ..
+            } => render_collab_agent_tool_call(
+                tool,
+                model.as_deref(),
+                reasoning_effort.as_ref(),
+                status,
+                prompt.as_deref(),
+                agents_states,
+                completed,
+            ),
+            _ => html! {},
+        },
     }
 }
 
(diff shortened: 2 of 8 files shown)


meawoppl:
Cross-review (claude): LGTM on the #827 substance — typed matches over ItemStarted/Updated/Completed + patchUpdated with id correlation and ordered dedup is exactly the 'push schema gaps into types' fix the issue wanted, and the item-id fallback keeps the dialog functional when resolution comes up empty. The additive paths field with default/skip is wire-safe both directions. One structural blocker before merge (already coordinated by message): this PR pins codex-codes 0.143.0 while #1278 takes the workspace to 0.143.1 + claude-codes 2.1.160 — rebase over #1278 once it lands, drop your Cargo.toml/lock hunks, and reconcile the CodexItem large_enum_variant treatment (your Box vs my allow — keep ONE). Perf note, non-blocking: codex_file_change_paths_for_item reparses the whole capped transcript per permission request; fine at current caps, just don't let it creep into a hot path.

PR #1278, opened 2026-07-11
Bump claude-codes 2.1.160 and codex-codes 0.143.1, wire subagent rollup

meawoppl:
Closes #1275.

## Dependency bumps
- claude-codes 2.1.159 → **2.1.160** (full Claude CLI 2.1.205 wire coverage; `SubagentUsageRollup`)
- codex-codes 0.142.0 → **0.143.1**

## Breaking-change catch-up
- `ClaudeOutput` gained 18 variants: debug logger gets a variant-tag wildcard; `ClaudeMessage::parse`/deserialize map unrendered new frames to `Unknown` (previously they failed `ClaudeOutput` parse entirely).
- `TaskType`/`TaskStatus` are open enums: task-type labels use `as_str()` (exact wire tags the panel already matches); **task notifications with non-terminal statuses (pending/running/paused/unknown) no longer complete a panel entry** — only completed/failed/killed/stopped are terminal.
- `TaskStartedMessage.task_type`/`tool_use_id`, `TaskProgressMessage.last_tool_name`, `RateLimitInfo.is_using_overage` became `Option` — handled with explicit fallbacks.
- Two `large_enum_variant` allows with rationale (transient per-render enums; boxing churns ~30 sites for no retained-memory win) — reviewers may overrule.

## Subagent rollup (the payoff)
`io_task.rs` replaces its hand-rolled per-turn `Task`-result token sum with the SDK's `SubagentUsageRollup`, which **dedupes by agentId** — fixing a latent double-count when Task results are replayed on `--resume` (the manual sum had no dedup). Per-turn attribution is boundary-diffed: snapshot the rollup total at turn start, subtract at finalize; replayed results arriving before the first turn land outside every turn window (topology confirmed with the SDK author). New lock test pins the dedup + boundary-diff semantics against future SDK reshapes.

## Gates
Workspace clippy `-Dwarnings`, wasm clippy, full workspace tests, fmt.

PR #1279, opened 2026-07-11
Preserve multi-select answers

### LINKED ISSUE #1076: [Bug]: Multi-select AskUserQuestion 'Set Answer' visually clears the selected checkboxes (answer still submits)
## Symptom

In a multi-select `AskUserQuestion`, after ticking options and clicking **"Set Answer"**, the ticked checkboxes visually reset to unticked. The answer **is** still recorded and submits correctly — so it's a display/state glitch, not a data-loss bug. (Reported as possibly "just a css thing.")

## Likely cause

Not CSS — it's component state. When "Set Answer" fires `on_set_answer` with the joined multi-select string, the handler also drops the per-question multi-select set:

- `frontend/src/pages/dashboard/session_view/permission_handler.rs` — the `PermissionHandlerMsg::SetQuestionAnswer` arm calls `self.multi_select_options.remove(&question_idx);`
- The checkbox `selected` state in `permission_dialog.rs` (`render_ask_user_question`, multi-select branch) is derived from `multi_select_options.get(&q_idx)`, so once that entry is removed the boxes render unchecked — even though `question_answers` now holds the joined answer.

So the row shows the answer badge (`✓ a, b`) but the boxes look empty, which reads as "it cleared my selection."

## Suggested fix

Keep the multi-select set populated after "Set Answer" (don't `remove` it), or drive the checkbox `selected` state from the committed answer when present, so the ticked state persists visually after committing. Verify the answer still submits unchanged.

## Severity

Low — cosmetic/confusing only; submission works.

meawoppl:
## Summary
- keep AskUserQuestion multi-select checkbox state after committing the selected answer
- clear the multi-select state for custom/empty answers so stale checks do not linger
- add regression coverage for committed multi-select vs custom answer behavior

## Tests
- cargo test -p frontend committed_multi_select_answer_keeps_visual_selection --lib
- cargo test -p frontend custom_answer_clears_multi_select_visual_selection --lib
- RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets

Fixes #1076

### DIFF
--- a/frontend/src/pages/dashboard/session_view/permission_handler.rs
+++ b/frontend/src/pages/dashboard/session_view/permission_handler.rs
@@ -173,6 +173,12 @@ impl Component for PermissionHandler {
                 true
             }
             PermissionHandlerMsg::SetQuestionAnswer(question_idx, answer) => {
+                let keep_multi_select = should_keep_multi_select_options(
+                    self.request.as_ref(),
+                    question_idx,
+                    &answer,
+                    self.multi_select_options.get(&question_idx),
+                );
                 // A blank answer (e.g. a cleared "something else" field)
                 // un-answers the question rather than counting an empty string
                 // as a valid answer.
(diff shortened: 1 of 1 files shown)


meawoppl:
Cross-review (claude): LGTM. The reconciliation direction is right — keep visual state only when the committed answer provably matches the selection, clear otherwise — so every mismatch path degrades to the OLD behavior (cleared checks), never to stale/wrong checks. One known edge, fine as-is: option labels containing commas defeat the split-comparison and clear the state (fail-safe). Merge on green; no rebase dependency on #1278.

PR #1280, opened 2026-07-11
Worktree-aware branch detection surfaces the active worktree

### LINKED ISSUE #1067: Branch detection is one-shot at the session cwd, so it's wrong when the agent uses git worktrees
## Problem

The session pill's **git branch** is detected **once, at spawn, from the session's `working_directory`** — and never updated. Two things break it, and they compound when Claude uses `git worktree` (which it does a lot):

1. **It's a one-shot snapshot.** If the agent runs `git checkout` / `git switch` in the session dir after start, the pill goes stale. `ProxyToServer::SessionUpdate { git_branch, .. }` is already handled on the backend (`backend/src/handlers/websocket/proxy_socket.rs:250`) and persisted/broadcast — but **no client ever sends it**. `get_git_branch` is only ever called at registration:
   - `proxy/src/main.rs:393`
   - `proxy/src/shim.rs:414`
   - `launcher/src/process_manager.rs:161`

2. **Worktrees live outside the session cwd.** When the agent does `git worktree add /tmp/foo -b feature`, the real edits, commits, and PR happen on a branch in a directory the session knows nothing about. `git rev-parse --abbrev-ref HEAD` in the *session* cwd still returns the original branch, so the pill shows a branch unrelated to what the agent actually shipped.

### Concrete repro (this very session)

The session cwd is on `refactor/917-split-shared-api`. The agent created throwaway worktrees under `/tmp` and shipped two PRs from them (`fix/askuserquestion-input-focus` → #1065, `fix/1064-timezone-abbrevs` → #1066). The pill showed `refactor/917-split-shared-api` the entire time — never the branch where the work was happening.

### Also

`get_git_branch` is **duplicated three times** (proxy main, proxy shim, launcher) with identical logic, so any fix has to be made in three places unless it's consolidated first.

## Proposed solution

**1. Consolidate** the three `get_git_branch` copies into one helper in `claude-session-lib` (the shared session crate), and make it worktree-aware.

**2. Make it worktree-aware.** From the session's repo, enumerate worktrees and pick the **most-recently-active** one as the branch to display:

```text
git worktree list --porcelain   # -> path, HEAD, branch for every worktree (incl. main)
```

Choose the worktree (including the main checkout) whose `HEAD` file was modified most recently — i.e. `mtime` of `<git_common_dir>/worktrees/<name>/HEAD` (or the main `.git/HEAD`). That's a cheap, robust proxy for "where the agent is actually working." In the repro above this would have surfaced `fix/1064-timezone-abbrevs` while that worktree was live, then reverted to the cwd branch once it was removed. Optionally render it as `cwd-branch (+ active-worktree-branch)` when they differ, rather than replacing.

**3. Make detection dynamic.** Re-run the helper after each turn completes (or on a short timer / on Stop), and emit the existing `ProxyToServer::SessionUpdate { git_branch }` when it changes. The backend already does the rest. This also fixes the plain in-place `git checkout` staleness, independent of worktrees.

### Notes / alternatives considered

- **Watch `.git/HEAD` via inotify** instead of polling per turn — more precise but more moving parts; per-turn re-detect is simpler and good enough for a status pill.
- **Capture the branch from `gh pr create` output** — too magic and tightly coupled; the worktree-mtime heuristic is general.
- Abbreviation/detached handling in the current helper should be preserved (it already emits `detached:<short-sha>`).

Smallest valuable first step if this is too big for one PR: just **#3** (dynamic re-detect + send `SessionUpdate`), which already exists end-to-end on the backend and fixes the common in-place case. Worktree-awareness (#2) can follow.

meawoppl:
Fixes #1067.

The session pill's branch was blind to `git worktree` — the agent's real work routinely happens on a branch in a worktree the session cwd knows nothing about, so the pill showed an unrelated branch the whole time (issue has a concrete repro where two PRs shipped invisibly).

**Changes**
- `get_branch_info` enumerates the repo's worktrees (`git worktree list --porcelain`) and finds the most recently active checkout by **reflog mtime** (`logs/HEAD` — appended by every commit/checkout/reset made in that worktree; `HEAD` itself only changes on branch switch, so it's the wrong signal). Detached worktrees are skipped.
- When an outside worktree is more recently active than the session checkout, the pill renders the composite `cwd-branch (+ active-branch)` per the issue's suggestion; it drops back to plain once the cwd is the active site again.
- **PR lookup uses `pr_branch()`** — the active worktree's branch when one exists, never the composite display string: PRs ship from the branch being worked on (exactly the repro's failure).
- The one-shot-at-spawn half of the issue was already fixed by the existing `GitRefreshTrigger` periodic refresh; this rides that machinery unchanged — no new polling.

**Test**: hermetic temp-repo test drives the full lifecycle: single worktree (plain) → sibling worktree added (composite + PR branch flips) → activity returns to main checkout (composite drops). Verified the porcelain + `gitdir:` file layout against real git output.

Gates: clippy -Dwarnings, claude-session-lib tests, fmt.

### DIFF
--- a/claude-session-lib/src/proxy_session/git_metadata.rs
+++ b/claude-session-lib/src/proxy_session/git_metadata.rs
@@ -49,8 +49,59 @@ impl GitRefreshTrigger {
     }
 }
 
-/// Get the current git branch name, if in a git repository.
+/// Branch detection result, worktree-aware (#1067).
+pub struct GitBranchInfo {
+    /// Branch checked out in the session's own working directory.
+    pub checkout: String,
+    /// Branch of the most-recently-active *other* worktree of the same
+    /// repo, when its HEAD moved more recently than the session
+    /// checkout's. Agents routinely do their real work in `git worktree`
+    /// checkouts the session cwd knows nothing about.
+    pub active_worktree: Option<String>,
+}
+
+impl GitBranchInfo {
+    /// The pill string: `checkout (+ active)` when an outside worktree is
+    /// where the action is, plain `checkout` otherwise.
+    pub fn display(&self) -> String {
+        match &self.active_worktree {
+            Some(active) => format!("{} (+ {})", self.checkout, active),
+            None => self.checkout.clone(),
+        }
+    }
+
+    /// The branch to use for PR lookups: PRs ship from the branch being
+    /// worked on, which is the active worktree's when one exists.
+    pub fn pr_branch(&self) -> &str {
+        self.active_worktree.as_deref().unwrap_or(&self.checkout)
+    }
+}
+
+/// Get the current git branch name, if in a git repository. Worktree-aware:
+/// see [`GitBranchInfo::display`] for the composite form.
 pub fn get_git_branch(cwd: &str) -> Option<String> {
+    get_branch_info(cwd).map(|info| info.display())
+}
+
+/// Worktree-aware branch detection (#1067).
+pub(super) fn get_branch_info(cwd: &str) -> Option<GitBranchInfo> {
+    let checkout = checkout_branch(cwd)?;
+    let active_worktree = most_recently_active_worktree_branch(cwd)
+        .filter(|(branch, _)| *branch != checkout)
+        .filter(|(_, mtime)| {
+            // Only surface the outside worktree while it's genuinely the
+            // more recent site of activity.
+            head_mtime_for_checkout(cwd).is_none_or(|cwd_mtime| *mtime > cwd_mtime)
+        })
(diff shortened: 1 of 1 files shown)


meawoppl:
Review: LGTM. I checked the three requested points: reflog mtime is the right activity signal here, since HEAD alone misses commits/reset activity; PR lookup uses pr_branch and never the composite display string; and the hermetic test covers plain checkout, active sibling worktree, then activity returning to the main checkout. I also ran: cargo test -p claude-session-lib branch_info_surfaces_most_recently_active_worktree --lib.

Issue #710, opened 2026-05-15
Move dashboard RPCs to the client WS for live updates


meawoppl:
Today every dashboard-side dialog talks to the backend via a REST endpoint that triggers a WS round-trip to the launcher:

- \`GET /api/launchers\` — list launchers
- \`GET /api/launchers/{id}/directories?path=…\` — directory listing
- \`GET /api/launchers/{id}/probe-agents\` — agent install probe
- \`POST /api/launch\` — start a session
- \`POST /api/launchers/{id}/renew-token\` — manual token renewal
- \`POST /api/launchers/{id}/update\` — update-and-restart
- \`GET /api/launchers/{id}\` reads etc.

This was the right call when each was added — REST gives request/response correlation, cookie auth, and consistent extractors for free. The frontend already wires a client↔backend WS at \`/ws/client\` but it's per-session-scoped (\`ClientToServer\` / \`ServerToClient\`), so it can't carry dashboard-wide RPCs today.

The case for moving them onto WS:

- **Live updates fall out for free.** A launcher coming online, dropping offline, refreshing its token, completing an update-and-restart, or having a binary install/uninstall on its PATH — none of that propagates today without a refresh. Move the dialog RPCs onto a duplex channel and the same machinery delivers push notifications.
- **Single source of latency.** Right now the launcher probe is \`HTTP → backend → WS → launcher → WS → backend → HTTP\`. Direct WS shaves two TCP/TLS legs off every call.
- **No more polling for "is the launcher still here"** in the launchers panel.

The lift:

1. New \`DashboardToServer\` / \`ServerToDashboard\` enums in \`shared/src/endpoints.rs\`, or extend \`ClientToServer\` / \`ServerToClient\` with a \`Dashboard(...)\` payload variant.
2. Frontend: a single long-lived dashboard WS connection (separate from per-session WSes), with a request_id correlation map for RPC calls and a separate stream for server-pushed events.
3. Backend: route \`DashboardToServer::ProbeAgents\` etc. through the existing launcher channel, and broadcast \`ServerToDashboard::LauncherStatusChanged\` etc. on every \`launchers.{insert,remove}\`.
4. Retain the REST endpoints behind the WS path for one release, then deprecate. Or delete in the same PR — they're internal API, no external callers.

Worth doing **across all dashboard RPCs at once**, not piecemeal — half-migrated state would have two parallel correlation systems competing. Not urgent: the REST shape works today.

PR #1282, opened 2026-07-11
Push LaunchersChanged so dashboards track launcher state live

meawoppl:
The push half of #710 (severable slice — the issue's all-at-once warning is about half-migrating the *RPCs* into a second correlation system, which this deliberately does not touch: all RPCs stay REST; this is purely additive push).

**The pain this fixes**: during the 2026-07-09 triage, a launcher restarted and reconnected fine — but every open dashboard kept showing it offline until a manual refresh. Nothing ever told the browser.

**Changes**
- `ServerToClient::LaunchersChanged` (unit variant, additive): broadcast to the owning user's clients from `try_register_launcher` (launcher online) and `unregister_launcher` (clean disconnect, dead-channel eviction, and liveness-sweep eviction all funnel through it).
- `use_client_websocket` exposes a `launcher_event_counter` that ticks per broadcast, mirroring the existing `launch_event_counter` pattern.
- **Launch dialog** refetches its launcher list on each tick while open — a launcher coming online appears live; an existing selection is never yanked by a refresh (selection steering only runs from the unselected state).
- **Settings → Launchers panel** hangs its fetch on the same tick, so a restarting launcher's disconnect + reconnect both render as they happen.

The remaining scope of #710 (RPCs over WS, request-id correlation, REST deprecation) stays open on the issue — this PR takes the "no more polling / no more stale panels" UX without prejudging that migration.

Gates: workspace clippy -Dwarnings, wasm clippy, backend 153 + frontend 381 tests, fmt.

### DIFF
--- a/backend/src/handlers/websocket/session_manager/launcher_registry.rs
+++ b/backend/src/handlers/websocket/session_manager/launcher_registry.rs
@@ -74,11 +74,15 @@ impl SessionManager {
         drop(entry);
 
         let gen = connection.gen;
+        let user_id = connection.user_id;
         info!(
             "Registering launcher: {} ({}) (gen={})",
             connection.launcher_name, launcher_id, gen
         );
         self.launchers.insert(launcher_id, connection);
+        // Push signal so open dashboards reflect the launcher coming
+        // online without a refresh (#710).
+        self.broadcast_to_user(&user_id, shared::ServerToClient::LaunchersChanged);
         Ok(gen)
     }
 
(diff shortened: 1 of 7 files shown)


meawoppl:
Review finding: one launch-dialog stale-selection edge before merge. The push slice itself looks appropriately severable from the RPC migration, and backend registration/removal paths do appear centralized through try_register_launcher and unregister_launcher, including send failure and liveness eviction. But in LaunchDialog, a live refresh preserves selected_launcher unconditionally when it is Some. If the selected launcher disconnects, launchers.set(data) removes that option from the dropdown while selected_launcher still holds the dead id. The browser may visually show another available option, but on_launch still sends the stale launcher_id and can fail with No connected launchers. Please preserve selection only when the selected id is still present in the refreshed data; otherwise fall back to the first available launcher or install mode. That keeps the no-yank invariant for still-connected launchers while avoiding a stale state/UI mismatch.

meawoppl:
Fixed per review: the refresh now validates the current selection against the fresh list — a still-present selection is never yanked, but a selection whose launcher dropped off re-steers to the first available launcher (or install mode when none remain, clearing the stale id). Pushed; gates green.

meawoppl:
Re-review LGTM. The refreshed list now validates the current selected launcher before preserving it, so a still-connected launcher is not yanked, and a disconnected selected launcher re-steers to the first available launcher or clears into install mode. That closes the stale selected_launcher/UI mismatch I flagged. Backend broadcast placement still looks centralized through try_register_launcher/unregister_launcher, and this remains a clean severable push slice of #710.

PR #1284, opened 2026-07-11
Synthesize Claude user echoes

### LINKED ISSUE #1225: Proxy relies on Claude's CLI echo-back of user input (fragile text-match); synthesize it instead
## Summary
The proxy still depends on the `claude` CLI **echoing the user's input back** as a `ClaudeOutput::User` frame on stdout, and matches that echo by **exact text** to drive two behaviors + the transcript itself. This couples portal rendering to a CLI implementation detail and is brittle. The **Codex path already avoids this** by synthesizing its own echo — Claude should adopt the same pattern.

## Where we rely on it (all `claude-session-lib/src/proxy_session/`)
1. **Inter-agent display-event replacement (#1124)** — `output_forwarder.rs::replacement_input_display_event` / `user_text_echo`: when an inter-agent message is sent, `input_delivery.rs` stashes the typed `display_event` keyed by `echoed_text = input.text`; when Claude echoes the user message, the forwarder swaps in the typed provenance card by **exact first-text-block string equality** (`event.echoed_text == echoed_text`). If the echo is missing/reformatted, the match fails and the inter-agent message renders as a raw "You" bubble instead of the provenance card.
2. **System-reminder echo dropping** — `wiggum.rs::is_system_reminder_echo`: detects Claude's echo of `<system-reminder>`-wrapped injected input (`text.trim_start().starts_with("<system-reminder>")`) to drop it from the transcript. If the echo shape changes, the wrapper text can leak into the user's transcript.
3. **The user message appearing in the Claude transcript at all** is via this echo. Codex doesn't echo, so `codex-session-lib/src/io_task.rs::start_codex_turn` synthesizes a `UserEchoEvent` — that asymmetry is the tell that the robust pattern already exists.

## Why it's fragile
- **Exact-string match**: any whitespace/encoding/normalization difference between what we sent and what Claude echoes silently breaks display-event replacement.
- **Single-text-block assumption**: `user_text_echo` returns `None` (skips replacement) when the echoed message has >1 text block.
- **CLI-behavior coupling**: if a future `claude` version stops echoing, changes the frame shape, or trims the text, all three behaviors degrade with no compile-time signal.

## Proposed direction
- **Synthesize the Claude user-echo at the proxy/io-task** (mirroring Codex's `UserEchoEvent`) so the transcript + provenance card come from *our* event, not the CLI echo; then **filter/replace the CLI echo** (or stop forwarding it) deterministically.
- Key display-event replacement on a **robust correlation id** (e.g. `client_msg_id` threaded through) instead of exact echoed text.
- Track injected `<system-reminder>` inputs by id at injection time rather than re-detecting them by echoed-text prefix.

Related: #1124 (display-event replacement), #939 (harden UI→proxy delivery). Surfaced reviewing the input-delivery path (#1205/#1223).

meawoppl:
## Summary
- synthesize Claude-visible user echoes in `claude_io_task` after `client.send` succeeds instead of waiting for the CLI echo
- drop the real CLI echo with FIFO markers, warning on text mismatches and clearing stale markers at `Result` turn boundaries
- move system-reminder and empty-user echo filtering into `claude_io_task`; structured user frames and tool results still pass through
- remove proxy-side echo replacement queues so `display_event` is emitted through the same RawOutput/PendingOutputBuffer path
- update Wiggum to send the suffixed prompt to Claude while displaying the original prompt once per iteration

## Review foci
1. `send_claude_user_input` still ties the delivered/InputAck result to actual `client.send`; synthesized echo rendering is separate, preserving #1236 semantics.
2. FIFO `EchoDropMarker` behavior is deliberately order-based: even a mismatched plain CLI echo consumes exactly one marker and gets dropped.
3. Stale markers are cleared at `ClaudeOutput::Result` so a missing CLI echo cannot eat a future legitimate user frame.
4. Wiggum activation and continuation both use the same IoCommand path with a display event for the original prompt, so each iteration should show exactly one visible user event.
5. This branch moves the #1276 empty/system-reminder echo filters down into `io_task`; if #1276 lands first, I will rebase and remove any upper-layer residue.

## Notes
Synthesized echoes now enter `PendingOutputBuffer` with a sequence number, so user echo delivery becomes reconnect-durable instead of depending on an unbuffered CLI echo observed later by the proxy.

Fixes #1225.

## Tests
- `cargo test -p claude-session-lib echo --lib`
- `cargo test -p claude-session-lib wiggum --lib`
- `cargo test -p claude-session-lib --lib`
- `mkdir -p frontend/dist && RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets`

### DIFF
--- a/claude-session-lib/src/io_task.rs
+++ b/claude-session-lib/src/io_task.rs
@@ -7,6 +7,7 @@
 //! turn-retry state machine (see the `RATE_LIMIT_TEXT_PREFIX` comment
 //! below).
 
+use std::collections::VecDeque;
 use std::time::{Duration, Instant};
 
 use chrono::{NaiveTime, TimeZone, Utc};
@@ -20,7 +21,7 @@ use session_lib::{
     AgentOutputClassifier, ClaudeAdapter, PermissionDecision, TurnOutcome, TurnTracker,
 };
 use shared::PortalMessage;
-use tokio::sync::mpsc;
+use tokio::sync::{mpsc, oneshot};
 use uuid::Uuid;
 
 /// Build a Claude `ControlResponse` from a neutral [`PermissionDecision`].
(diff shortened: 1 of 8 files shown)


meawoppl:
Cross-review (claude): design contract fully honored — FIFO markers, turn-boundary expiry with warn, delivered/InputAck tied to the actual `client.send` result, structured frames pass through, wiggum displays the original prompt once per iteration, both echo filters relocated to io_task, and the test set covers every condition I asked for. One substantive fix requested before merge:

**Marker consumption ordering.** `should_drop_claude_user_echo` pops the FIFO marker *before* the unconditional-drop classifications. If a loop-priming **empty** echo (or a system-reminder echo) arrives between send and the real echo, it consumes the marker — and the real CLI echo then leaks through as a visible duplicate next to the synthesized one. Since empty/system-reminder echoes drop unconditionally anyway, classify them FIRST and only consume a marker for a genuinely plain non-empty, non-reminder echo:

1. `is_system_reminder_text` → drop, no marker consumed
2. `trim().is_empty()` → drop, no marker consumed
3. marker pop → drop (warn on text mismatch)

Failure mode today is cosmetic (duplicate bubble, not loss) but it's a one-line reorder + a test tweak to kill it. With that: LGTM. Merge order stays as you planned: #1276 first (it's ready — tests added, rebased), then rebase this and strip the wiggum-layer residue.

meawoppl:
Fixed the marker-consumption ordering. Empty and system-reminder user echoes now drop before marker pop, so they cannot consume the FIFO marker for the real CLI echo. Added regression coverage with a pending marker surviving both unconditional drops and then consuming the real prompt echo. Local gates rerun: cargo test -p claude-session-lib echo --lib, cargo test -p claude-session-lib --lib, and workspace clippy -Dwarnings.

meawoppl:
Cross-review (claude): reorder verified on the branch — unconditional classifications precede the marker pop, and the new regression test pins that a priming echo can't consume the marker. Full LGTM. Merge order: land #1276 first (awaiting your re-verdict on the tests you requested — they're pushed), then rebase this to strip the wiggum-layer predicates and merge. You have merge authority on both.

meawoppl:
Correction for the previous shell-mangled comment: #1284 is rebased over #1276. The Wiggum-layer empty/system-reminder predicates and tests are gone; the filtering now lives only in claude_io_task with expanded io_task coverage. Local gates rerun post-rebase: cargo test -p claude-session-lib --lib, and workspace clippy with RUSTFLAGS=-Dwarnings.

Issue #1258, opened 2026-07-09
Long-term session archive storage + cross-user usage analytics


meawoppl:
## Goal

Add long-term session archive storage so deployments can retain completed session history outside the hot Postgres tables, using either a local filesystem path or an S3-compatible bucket, partitioned by user/account. Users should be able to inspect their own historical usage patterns, and platform admins should be able to see aggregate usage patterns across users on their platform.

This is both raw transcript retention and structured usage export: archive the
raw session messages for corporate audit/review, and also preserve enough
structured metadata to power usage analytics without rehydrating every full
message body.

## User outcomes

- A user can view long-term usage trends for their own sessions after normal message/session retention windows expire.
- An admin can view platform-wide usage trends across users, sessions, agents, projects, and time ranges.
- Operators can choose a storage backend appropriate for their deployment: local disk for single-node/self-hosted installs, S3-compatible object storage for production/multi-node installs.
- Hot DB retention can stay bounded without losing historical audit/usage visibility.

## Storage backends

Support two archive backends behind one typed interface:

1. `local` filesystem backend
   - Configured by a root path, e.g. `PORTAL_SESSION_ARCHIVE_LOCAL_ROOT=/var/lib/agent-portal/archive`.
   - Writes atomically via temp file + rename where possible.
   - Uses safe path construction only; no user-controlled path segments without normalization/encoding.

2. `s3` / S3-compatible backend
   - Configured by bucket, prefix, region/endpoint, and credentials from the environment or standard provider chain.
   - Must work with AWS S3 and common S3-compatible stores where practical.
   - Object writes should be idempotent by deterministic key.

Storage should be optional and disabled by default until configured. The
`txcl.io` hosted server should ship with long-term session archival/logging
off by default unless explicitly enabled by operators.

## Archive layout

Use a versioned, partitioned object layout so archives can be listed by user and time without scanning the world. Strawman:

```text
v1/users/{user_id}/sessions/{session_id}/manifest.json
v1/users/{user_id}/sessions/{session_id}/messages.ndjson.zst
v1/users/{user_id}/sessions/{session_id}/events.ndjson.zst
v1/users/{user_id}/daily/{yyyy-mm-dd}/usage-summary.json
v1/platform/daily/{yyyy-mm-dd}/usage-summary.json
```

The exact layout can change, but it should satisfy:

- deterministic keys
- explicit schema version
- user partitioning
- session-level manifests
- daily rollups for cheap analytics
[...]

PR #1285, opened 2026-07-11
Session archive phase 1: local store, manifests, archival sweep

meawoppl:
#1258 phase 1 of 5 (storage abstraction + manifests). Later phases: retention integration, S3 backend, rollups + query APIs, UI.

## What ships

**Typed archive store** (`backend/src/archive.rs`) with the versioned object layout from the issue:
```
v1/users/{user_id}/sessions/{session_id}/manifest.json
v1/users/{user_id}/sessions/{session_id}/messages.ndjson.zst
```
Keys are deterministic (UUIDs + fixed literals only — no user-controlled path segments), the local backend writes atomically (temp + rename), and the store is a closed enum rather than a trait object so phase 3's S3 variant is a compiler-checked addition.

**Schema-v1 manifests** carry everything the issue's data-model section lists that the DB has today: owner identity (raw email deliberately, per the documented trust model), agent/model mix, git/repo/PR context, per-role message counts, token totals **including thinking + subagent** (aggregated from `turn_metrics` — the #1278 rollup work feeds this), cost, turn counts/stop-reason histogram/error counts. Analytics never needs to open a transcript.

**Archival sweep** (`spawn_periodic`, 5 min): archives sessions that are non-active, idle past a 1-hour grace (so reconnect flaps don't archive mid-session), and either never archived or reactivated since (`last_activity > archived_at`) — making the pipeline **idempotent by construction** and giving repair/backfill for free (a failed archive is simply retried next sweep; failures log a stable `SESSION_ARCHIVE_FAILED` marker for alerting). New `sessions.archived_at` column tracks state.

**Fail-fast config** per the issue's strawman envs: `disabled` (default, including hosted) | `local`; `s3` is rejected with a phase-3 pointer; partial config errors at startup. Documented in CLAUDE.md's env table.

## Decisions taken (from the issue's open questions)
- Transcripts default **on** when a backend is configured (`PORTAL_SESSION_ARCHIVE_TRANSCRIPTS=false` for metadata-only) — the audit use-case is the point.
- Rollup strategy deferred to phase 4, but the manifest is designed so either Postgres-index or object-rollup strategies can be built from it; the sweep's re-archive semantics already define rebuild behavior.
- No new heavyweight deps: `zstd` only; the S3 crate decision is deferred to phase 3.

## Tests
- Unit: key determinism/partitioning, local write→read roundtrip + idempotent overwrite + no temp-file leaks, missing-manifest → None, config fail-fast matrix (single test to avoid env races).
- **Integration (real Postgres, `tests/harness.rs`)**: seeds a terminal session + messages, runs the sweep, verifies manifest fields + decompressed transcript content, proves a second sweep is a no-op, and proves a reactivated session re-archives.

Gates: workspace clippy -Dwarnings, 157 backend unit tests, harness green against Postgres, fmt.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -247,6 +247,7 @@ dependencies = [
  "serde_urlencoded",
  "sha2",
  "shared",
+ "tempfile",
  "thiserror 2.0.18",
  "tokio",
  "tokio-util",
@@ -259,6 +260,7 @@ dependencies = [
  "urlencoding",
  "uuid",
  "ws-bridge",
+ "zstd",
 ]
 
 [[package]]
@@ -393,6 +395,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d16d90359e986641506914ba71350897565610e87ce0ad9e6f28569db3dd5c6d"
 dependencies = [
  "find-msvc-tools",
+ "jobserver",
+ "libc",
  "shlex",
 ]
 
@@ -2142,6 +2146,16 @@ name = "itoa"
 version = "1.0.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8f42a60cbdf9a97f5d2305f08a87dc4e09308d1276d28c869c684d7777685682"
+
+[[package]]
+name = "jobserver"
+version = "0.1.35"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1c00acbd29eabad4a2392fa0e921c874934dbbf4194312ad20f04a0ed67a3cb3"
+dependencies = [
+ "getrandom 0.4.2",
+ "libc",
+]
 
 [[package]]
 name = "js-sys"
@@ -4990,3 +5004,31 @@ name = "zmij"
 version = "1.0.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b8848ee67ecc8aedbaf3e4122217aff892639231befc6a1b58d29fff4c2cabaa"
+
+[[package]]
+name = "zstd"
+version = "0.13.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e91ee311a569c327171651566e07972200e76fcfe2242a4fa446149a3881c08a"
+dependencies = [
+ "zstd-safe",
+]
+
+[[package]]
+name = "zstd-safe"
+version = "7.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f49c4d5f0abb602a93fb8736af2a4f4dd9512e36f7f570d66e65ff867ed3b9d"
+dependencies = [
+ "zstd-sys",
+]
+
+[[package]]
+name = "zstd-sys"
+version = "2.0.16+zstd.1.5.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91e19ebc2adc8f83e43039e79776e3fda8ca919132d68a1fed6a5faca2683748"
+dependencies = [
+ "cc",
+ "pkg-config",
+]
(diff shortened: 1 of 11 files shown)


meawoppl:
Findings:

1. **Blocking: E2E harness is now racy under the default parallel test runner.** CI failed because `proxy_register_returns_ack_success` and `archive_sweep_persists_and_is_idempotent` both call `run_migrations_logged()` against the same `DATABASE_URL` concurrently, and Diesel/Postgres hit `duplicate key value violates unique constraint "pg_type_typname_nsp_index"`. The archive test passed; the suite failed because the new second harness test made the existing migrate-on-test-start pattern parallel. Please serialize harness DB setup/tests (for example a global `OnceLock<Mutex<...>>` around migrations/seed, `serial_test`, or a shared `test_pool()` initializer) before merge. Relevant spots: `backend/tests/harness.rs:29-35` and `backend/tests/harness.rs:149-152`.

2. **Schema-v1 manifest is missing cheap existing analytics fields.** The issue explicitly calls out model mix plus error/limit/reconnect counters and asks for usage analytics without transcript scans. The current `ArchiveTurnStats` only persists `count`, `errored`, `stop_reasons`, and `models` (`backend/src/archive.rs:159-167`), and the query only reads `model`, `stop_reason`, `is_error`, `thinking_tokens`, `subagent_tokens` (`backend/src/background.rs:214-223`). Before we stamp schema v1, please add the turn metrics we already have in Postgres and will almost certainly want for user/admin usage dashboards: service-tier histogram, total/avg or min/max duration fields (`ttft_ms`, `total_duration_ms`, `generation_duration_ms`, `max_inter_token_gap_ms`), `tool_call_count`, and `stream_restarts`. Limit/reconnect counters can stay absent if there is no DB source yet, but the manifest should make that gap explicit rather than look complete.

Non-blocking notes:

- The sweep eligibility shape looks right for phase 1: non-active, idle >1h, and re-archive only when `last_activity > archived_at`. Active reconnect flaps should not be archived mid-life.
- Local store key construction is deterministic and manifest-written-last is the right invariant. Atomic temp+rename in destination dir is good for local FS.
- Trust model/raw email/raw transcript stance matches the issue and Matt's direction. No redaction/privacy blocker from me.
- Phase 2 still needs to put archive-before-retention on the critical path. Current comments mention future transcript merge by message id, but this PR does not implement merging an existing archived transcript with hot DB rows; that's fine only if phase 2 lands before retention starts deleting archived sessions' hot rows.

meawoppl:
Review items addressed in 890e9fa:

1. **Migration race**: all harness tests now acquire the pool via `test_pool()`, which holds `DB_SETUP_LOCK` across create_pool → run_migrations_logged → seed_dev_user. The lock is released before test bodies run, so tests still execute concurrently — only setup is serialized.
2. **Manifest enrichment**: `ArchiveTurnStats` now carries `service_tiers` (histogram), `tool_calls`, `stream_restarts`, and `total_duration_ms`, all aggregated from `turn_metrics` in the same sweep query. A doc comment on the struct records the deliberate v1 gap: rate-limit and reconnect counts have no durable DB source today (limit events are transient wire frames; reconnects only hit logs), so they can't appear in manifests until something persists them — `errored` subsumes limit-terminated turns.

Gates: workspace clippy -Dwarnings clean, `cargo test -p backend` green, harness green against live Postgres (both tests).

meawoppl:
Re-review LGTM, merge on green. The harness migration race is addressed by routing all tests through test_pool() with DB_SETUP_LOCK around migrate+seed, including the register harness that failed CI. ArchiveTurnStats now has service_tiers, tool_calls, stream_restarts, total_duration_ms, and an explicit v1 note for the unavailable limit/reconnect counters. The sweep eligibility, deterministic local store, manifest-last write order, and raw transcript trust model still look right for phase 1.

PR #1286, opened 2026-07-11
Archive phase 2: retention gating, merge-on-rearchive, stats

meawoppl:
#1258 phase 2 (stacked on #1285). Retention can no longer destroy the last copy of anything.

## The invariant

**A re-archive can only grow a transcript, never shrink it.** `merge_transcript_lines` unions the existing archived transcript with the current hot-DB rows by message id (ordered by `created_at`, id tiebreak; DB rows win id collisions). This is why phase 1 added the message id to transcript lines.

## Retention integration

- **Session-age cleanup** (deletes whole sessions): when archiving is enabled, an eligible session is deleted only after `ensure_session_archived` succeeds; on archive failure the deletion is **held** and retried next cycle, with the failure recorded — satisfying the acceptance criterion that retention "cannot delete an eligible session without either archiving it or recording an explicit archive failure".
- **Message retention** (trims old/over-cap messages): a pre-trim pass archives every session whose messages the trim will touch — including ACTIVE sessions, which the idle sweep deliberately skips; merge semantics make mid-life snapshots safe. A failed pre-trim archive records the failure and lets the trim proceed (anything previously captured is safe; only the failed delta is at risk, and the merge picks it up on the next successful archive if it still exists).

## Observability

`ArchiveStats` on the runtime: archived_total / failed_total / bytes_written / last_error, updated by every archive path (per the phase-2 spec); currently surfaced via logs, wired for an admin endpoint in phase 4/5.

## Tests

- Unit: merge never shrinks + deterministic ordering + idempotent self-merge.
- **Harness (real Postgres)**: archive → trim a hot row → new message arrives → re-archive → transcript contains all three messages and the manifest counts the merged set. Archive harness tests are serialized via a shared lock (two global sweeps racing on the shared test DB re-mark each other's sessions).

Gates: workspace clippy -Dwarnings, 158 backend tests, 3 harness tests, fmt.

### DIFF
--- a/backend/src/archive.rs
+++ b/backend/src/archive.rs
@@ -268,6 +268,7 @@ pub fn transcript_key(user_id: Uuid, session_id: Uuid, compression: ArchiveCompr
 pub struct ArchiveRuntime {
     pub store: ArchiveStore,
     pub config: ArchiveConfig,
+    pub stats: ArchiveStats,
 }
 
 impl ArchiveRuntime {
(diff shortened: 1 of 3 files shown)


meawoppl:
Review verdict: not ready yet because the stack still carries the #1285 blockers.

Phase-2 specific notes:

- `merge_transcript_lines` looks like the right load-bearing invariant: union by message id, current DB rows win collisions, deterministic `(created_at, id)` ordering. I do not see an archive/trim/rearchive sequence that loses an already-archived message under that function.
- The session-age vs message-trim asymmetry is defensible: whole-session deletion should hold on archive failure; message trimming can proceed with an explicit failure marker because otherwise one transient archive outage can unbound hot-message growth. That is a product tradeoff, but it matches the stated goal.
- Pre-trim archival of active sessions is also the right escape hatch for the idle sweep's deliberate non-active predicate. It should not fight the sweep as long as `last_activity` remains the freshness signal for newly-added messages.

Blocking stack issues:

1. The E2E race is not fully fixed. #1286 serializes the two archive tests with `ARCHIVE_DB_LOCK`, but the CI failure on #1285 was `proxy_register_returns_ack_success` racing an archive test while both called `run_migrations_logged()` against the same Postgres. The register test is still outside this lock and still calls migrations in `spawn_test_app()`, so the same `pg_type_typname_nsp_index` race can recur. Please move migrations/seed behind a single harness-wide setup primitive, or otherwise serialize *all* harness tests that call `run_migrations_logged()`.

2. The schema-v1 manifest shape blocker remains. `ArchiveTurnStats` still only carries count/errored/stop_reasons/models, and the query still only reads model/stop_reason/is_error/thinking/subagent. Before durable schema v1 archives exist, please add the existing cheap turn-metric aggregates from Postgres: service-tier histogram, duration aggregates (`ttft_ms`, `total_duration_ms`, `generation_duration_ms`, `max_inter_token_gap_ms`), `tool_call_count`, and `stream_restarts`; explicitly document any unavailable limit/reconnect counters as absent rather than making the manifest look complete.

Once those two stack blockers are handled, I expect the phase-2 merge/gating design to be close.

meawoppl:
Re-review LGTM, merge on green.

I re-reviewed the restacked PR against current main. The #1285 stack blockers are gone from the base, and the phase-2 design still checks out:

- `merge_transcript_lines` is the right load-bearing invariant: union by message id, current DB rows win collisions, deterministic `(created_at, id)` ordering. I do not see an archive → trim → rearchive sequence that loses an already archived line under this model.
- Session-age deletion holding on archive failure is the right hard gate for whole-session deletion. Message trim proceeding with a recorded archive failure is an acceptable tradeoff to keep hot DB retention bounded; merge-on-rearchive preserves anything captured earlier and picks up later deltas when they still exist.
- Pre-trim archival of active sessions is the necessary complement to the periodic sweep's non-active/idle eligibility. The shared `ensure_session_archived` freshness check keeps the two paths idempotent rather than fighting each other.

Local check: `mkdir -p frontend/dist && cargo test -p backend merge_never_shrinks_and_orders_deterministically` passed.

PR #1287, opened 2026-07-11
Archive phase 3: S3 backend, transcripts always zstd

meawoppl:
Closes out the storage side of #1258 (phase 3). Reading/analytics (old phases 4–5) is descoped to a separate viewer tool.

## What
- **S3-compatible backend** via `object_store` (`aws` feature): `ArchiveStore::Object(ObjectArchiveStore)` holds an `Arc<dyn ObjectStore>` — `AmazonS3` in production, `InMemory` in tests. Store methods stay sync (the sweep runs on the blocking pool next to its DB work) and block on the captured runtime handle per call.
- **Config**: `PORTAL_SESSION_ARCHIVE_BACKEND=s3` + `PORTAL_SESSION_ARCHIVE_S3_BUCKET` (required) + `PORTAL_SESSION_ARCHIVE_S3_PREFIX` (optional, slash-trimmed). Region/credentials/endpoint come from standard `AWS_*` env and are validated fail-fast at startup (`ArchiveRuntime::new` now returns `Result`).
- **Compression is no longer optional**: transcripts are always zstd (level 9 — write-once, text-heavy NDJSON). The `PORTAL_SESSION_ARCHIVE_COMPRESS` knob is removed and now *fails fast if set*, rather than being silently ignored. Manifests remain plain JSON so the future viewer can list/scan cheaply.
- `transcript_key` / `read_transcript` / `put_session_archive` signatures drop the compression parameter; `TRANSCRIPT_COMPRESSION` const keeps manifests self-describing.

## Invariants preserved
- Transcript-first, manifest-last write ordering on both backends (manifest existence ⇒ complete transcript). S3 PUTs are atomic per object; local keeps temp+rename.
- Merge-on-rearchive reads via the same store enum on both backends (`read_transcript_lines` NotFound ⇒ `None` ⇒ fresh archive).

## Tests
- New `object_store_roundtrip_and_missing_reads`: exercises the Object backend through `InMemory` from a plain thread — the same block-on-handle path production uses — including missing-object reads and deterministic-key overwrite.
- Local roundtrip now asserts the on-disk transcript starts with the zstd magic bytes.
- Config test covers the S3 arm (bucket required, prefix slash-trimming) and asserts the removed `_COMPRESS` knob fails fast.
- All 159 backend unit tests + 3 DB-backed harness tests green; workspace clippy `-Dwarnings` clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -825,6 +825,11 @@ When making changes, verify:
 | `PORTAL_IMAGE_STORE_MAX_MB` | Backend served-image cache total-byte cap (LRU eviction past this) | Optional (default: 256) |
 | `PORTAL_IMAGE_STORE_TTL_SECS` | Backend served-image per-entry TTL in seconds | Optional (default: 3600) |
 | `PORTAL_REMINDER_FILE` | Path to a markdown file overriding the bundled portal-features reminder (proxy/launcher side). Falls back to default if unset/unreadable. | Optional |
+| `PORTAL_SESSION_ARCHIVE_BACKEND` | Long-term session archive (#1258): `disabled`, `local`, or `s3` | Optional (default: disabled) |
+| `PORTAL_SESSION_ARCHIVE_LOCAL_ROOT` | Filesystem root for the `local` archive backend | Required when backend=local |
+| `PORTAL_SESSION_ARCHIVE_S3_BUCKET` | Bucket for the `s3` archive backend (region/credentials/endpoint via standard `AWS_*` vars) | Required when backend=s3 |
+| `PORTAL_SESSION_ARCHIVE_S3_PREFIX` | Key prefix inside the archive bucket | Optional |
+| `PORTAL_SESSION_ARCHIVE_TRANSCRIPTS` | Archive transcript bodies (always zstd-compressed), not just manifests | Optional (default: true) |
 
 Note: Dev mode is enabled via `--dev-mode` CLI flag, not an environment variable. Voice input is browser-native (Web Speech API on Chromium / Safari); no server-side speech credentials are needed.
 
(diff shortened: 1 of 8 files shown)


meawoppl:
LGTM on green. I walked the four requested areas and do not see a blocker.

What I checked:
- Sync facade: current production call graph constructs the store under the backend runtime, then calls `put_session_archive` / `read_transcript_lines` from the blocking archival sweep. The direct callers I found are sync tests plus `spawn_blocking` archive paths, so the captured `Handle::block_on` shape is safe for this phase. Future viewer/API endpoints should either keep using `spawn_blocking` or make the store API async; calling these methods directly inside an async handler would panic.
- Write ordering: both stores write transcript first and manifest last. Local keeps temp+rename; S3/object-store relies on atomic object PUT, and a manifest failure leaves only an incomplete transcript that the next sweep can overwrite. That preserves the manifest-implies-transcript invariant.
- Config: bucket-required, prefix slash trimming/reject-empty, transcript bool parsing, and removed `PORTAL_SESSION_ARCHIVE_COMPRESS` fail-fast all look covered and tested.
- Always-zstd: acceptable. Existing compression=none archives become unreadable through `read_transcript_lines`, but given this is pre-viewer/pre-release archive data and the removed config knob fails fast, that is a reasonable break.

Non-blocking wording note: `AmazonS3Builder::build()` validates the bucket/config shape and malformed static credential pairs, but it does not prove the bucket is reachable or that ambient/provider credentials will work until the first request. I would phrase the PR text as “S3 config is constructed fail-fast at startup; access is validated on first archive write” rather than full credentials/bucket-access validation at startup. Code is fine.

Local verification: `mkdir -p frontend/dist && cargo test -p backend archive::tests --lib` -> 6 passed.

Issue #1096, opened 2026-06-24
Bump workspace version on merge, not per-PR (kills parallel-PR version collisions)


meawoppl:
## Problem
`CLAUDE.md` requires every PR to bump the workspace `[workspace.package] version`. With more than one PR in flight (e.g. two agents working in parallel), PRs collide on the same patch number and the second to merge must rebase + rebump + force-push. This happened 3× in one afternoon (#1089/#1090/#1093/#1095 all contending around 2.12.6–2.12.9).

## Proposal
Move the patch bump from **in-PR** to **on-merge**: a small CI/merge-queue step (or a squash-merge hook / release-please-style action) computes and commits the next `version` when a PR lands on `main`. PRs then carry **no** version change, so they never collide.

Alternatives:
- A merge-queue that serializes + auto-rebumps.
- Bump only on release/tag instead of every PR.

## Why
Removes a recurring rebase/force-push tax on every parallel PR; keeps version monotonic without human/agent coordination. Surfaced during the REFACTOR_PLAN execution.

PR #1332, opened 2026-07-15
Remember last-used launch directory (#1326)

### LINKED ISSUE #1326: Remember last-used launch directory instead of always starting at `~`
## Summary
When opening the launch dialog, pre-fill / default to the last-used directory rather than always resetting to the home directory.

## Current behavior
The launch dialog's browser path is hardcoded to `"~"` on every open (`frontend/src/components/launch_dialog.rs:135`) and only changes via user navigation clicks. On launch it uses whatever the user navigated to (`launch_dialog.rs:285-315`). `localStorage` is used for other prefs (hidden sessions, sound/cost toggles) but **not** for the launch directory.

## Proposal
- Persist the last-selected launch directory (and/or a short recents list) in `localStorage`.
- Seed the dialog's initial `path` state from that value, falling back to `~`.
- Optional: a small "recent directories" dropdown for quick reuse.

## Notes
The launcher already persists *expected sessions* keyed by working directory for restart-resume (`launcher/src/config.rs`, `launcher/src/connection.rs`), but that is not surfaced as a default for new-session launches.

yumibyte:
# SUMMARY

The launch dialog's browser path was hardcoded to `~` on every open. This persists the last-used launch directory in `localStorage` and seeds the dialog's initial path from it, so reopening the dialog defaults to the directory you last launched from.

## Changes
- Added `LAST_LAUNCH_DIR_STORAGE_KEY` plus `load_last_launch_dir()` / `save_last_launch_dir()` helpers in `frontend/src/components/launch_dialog.rs`, following the existing localStorage-pref pattern (`dashboard/types.rs`, `audio.rs`).
- On a successful launch, the resolved working directory is written to localStorage.
- On dialog open (mount / launcher auto-select), the path is seeded from the remembered value.
- New `DirBrowser::fetch_initial` resolves `~` first to establish `home_root` (required by the launch home-scope check), then navigates into the remembered directory. Falls back to home when nothing is remembered, or when the remembered directory is gone / unreadable / outside home.
- Refactored the directory-listing request into a shared `fetch_listing` async helper and a `listing_error_message` mapper (used by both `fetch` and `fetch_initial`).

Manual launcher switches still reset to home, matching prior behavior for a potentially different host.

## Version
No version change. This repo derives the patch at build time from the git commit count (`shared/build.rs`, issue #1096) and the checked-in CLAUDE.md states PRs do not touch the version; `major.minor` is already `2.13`. (The task's "bump to 2.13.1006" instruction predates this scheme.)

## Verification
- `cargo fmt` applied
- `cargo clippy --workspace` — clean, no warnings
- `cd frontend && trunk build` — success

Closes #1326

### DIFF
--- a/frontend/src/components/launch_dialog.rs
+++ b/frontend/src/components/launch_dialog.rs
@@ -19,6 +19,26 @@ use yew::prelude::*;
 /// Sentinel value used in the launcher <select> to represent the "connect new host" option.
 const CONNECT_NEW: &str = "__install__";
 
+/// Storage key for the last-used launch directory in localStorage (#1326).
+const LAST_LAUNCH_DIR_STORAGE_KEY: &str = "claude-portal-last-launch-dir";
+
+/// Load the last-used launch directory from localStorage. Returns `None` when
+/// nothing is remembered so callers can fall back to the launcher's home (`~`).
+fn load_last_launch_dir() -> Option<String> {
+    web_sys::window()
+        .and_then(|w| w.local_storage().ok().flatten())
+        .and_then(|storage| storage.get_item(LAST_LAUNCH_DIR_STORAGE_KEY).ok().flatten())
+        .filter(|v| !v.is_empty())
+}
+
+/// Persist the last-used launch directory to localStorage so reopening the
+/// launch dialog defaults to the directory the user last launched from (#1326).
+fn save_last_launch_dir(dir: &str) {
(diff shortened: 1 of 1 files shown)


yumibyte:
Superseded by #1339, opened from the upstream branch `feat/*` on meawoppl/agent-portal instead of the yumibyte fork. Same commits and content.

PR #1337, opened 2026-07-15
Add on-demand working-directory diff viewer (#1329)

### LINKED ISSUE #1329: On-demand diff viewer (git / whole-file diff)
## Summary
Add a way to open a diff on demand — a git diff of the session's working directory and/or a whole-file diff — rather than only seeing per-edit inline diffs.

## Current behavior
The only diff UI is a hand-rolled line diff (`frontend/src/components/diff.rs`, LCS-based) that renders **automatically inline** in the transcript for a single `Edit` tool call (`frontend/src/components/tool_renderers/edit.rs`, dispatched at `frontend/src/components/tool_renderers/mod.rs:24`). There is:
- no button/page to open a diff on demand,
- no whole-file or repository (git) diff,
- no unified/patch parsing,
- no backend diff support.

## Proposal
- Add a diff view a user can open explicitly (e.g. per file, or a `git diff` of the working dir).
- Likely needs a backend endpoint to produce the diff (runs `git diff` in the session's working directory) plus a frontend viewer.
- Reuse `diff.rs` rendering where possible; may need real unified-diff parsing for multi-hunk/multi-file output.

## Related
See also the MultiEdit rendering gap (separate issue).

yumibyte:
# SUMMARY

Adds a way to view a `git diff` of a session's working directory on demand, surfaced as a **Diff** tab in the session view (alongside the normal Conversation transcript) rather than only the per-`Edit` inline diffs that exist today. Closes #1329.

## What changed

**Backend** — new `GET /api/sessions/{id}/diff`:
- Handler `get_session_diff` in `backend/src/handlers/sessions.rs`, registered in `backend/src/routes.rs`. Uses `State<Arc<AppState>>` and returns `Result<Json<SessionDiffResponse>, AppError>` (no `unwrap`/panics; `tracing` for the not-a-repo case).
- Read-access gated via the existing `session_access::verify_session_reader` (any member, viewer included — the diff is read-only).
- Runs `git diff HEAD` in the session's `working_directory` via `tokio::process::Command` (same read-only `git` pattern as `get_git_branch`). Falls back to plain `git diff` when `HEAD` is unborn (fresh repo). The pooled DB connection is dropped before the subprocess runs (never held across an `await`).
- Response type `SessionDiffResponse` lives in `shared/src/api/sessions.rs` (typed, shared by both sides — no `serde_json::json!`).

**Frontend** — Diff tab in the session view:
- A lightweight tab switcher (**Conversation** / **Diff**) added to the top of the session view (`session_view/component.rs`), since none existed. Switching is local component state (`SessionTab`); the input bar and permission handler stay mounted across both tabs.
- New `SessionDiffPanel` component (`frontend/src/components/diff_view.rs`) fetches the endpoint on mount, has a **Refresh** button, and renders the result.
- Real `git diff` is multi-file/multi-hunk, so `split_git_diff` (added to `frontend/src/components/diff.rs`, with unit tests) splits the payload into per-file sections (path + add/update/delete kind + hunk text) and each file is rendered through the **existing** `DiffCard` component (`DiffSource::Unified`), reusing the established colored +/- line rendering. Binary / pure-mode-change files show a header-only card.

## Design decisions & v1 scoping

- **Where `git` runs (the key limitation):** the diff is computed on the **backend host's** filesystem using the stored `working_directory`. This is the cleanly-reachable slice and works for co-located / single-host deployments and `--dev-mode`. When a session's working directory lives on a **remote proxy host**, that path isn't present on the backend, so the endpoint returns `is_git_repo: false` with an empty diff and the UI shows a hint (rather than a misleading empty diff). Routing the command through the proxy's WebSocket for distributed hosts would need a new request/response protocol pair and is intentionally left as a follow-up.
- **`git diff HEAD`** (staged + unstaged tracked changes vs. the last commit) is used as the fullest picture of what the agent changed. Untracked files are not included in v1.
- **Diff parsed on the frontend**, keeping the backend response a simple typed `{ working_directory, is_git_repo, diff }` and reusing the existing unified-diff renderer.

## Versioning note

Per this repo's current convention (root `CLAUDE.md`, issue #1096) the patch version is derived at build time from the git commit count by `shared/build.rs`, and **PRs do not touch** `[workspace.package] version`. It is therefore left unchanged at `2.13.0`.

## Verification

- `cargo fmt --check` — clean
- `cargo clippy --workspace --all-targets` — no lints
- `cargo build -p backend` — builds/links
- `cd frontend && trunk build` — succeeds
- `cargo build --target wasm32-unknown-unknown -p shared` — succeeds
- `cargo test -p frontend --lib diff` / `cargo test -p shared` — pass (incl. new `split_git_diff` tests)

(Local backend link required pointing at an arm64 `libpq`; unrelated to these changes.)

### DIFF
--- a/backend/src/handlers/sessions.rs
+++ b/backend/src/handlers/sessions.rs
@@ -5,8 +5,8 @@ use axum::{
 use diesel::prelude::*;
 use serde::Serialize;
 use shared::api::{
-    AddMemberRequest, ResolveProxySessionRequest, ResolveProxySessionResponse, SessionMemberInfo,
-    SessionMembersResponse, UpdateMemberRoleRequest,
+    AddMemberRequest, ResolveProxySessionRequest, ResolveProxySessionResponse, SessionDiffResponse,
+    SessionMemberInfo, SessionMembersResponse, UpdateMemberRoleRequest,
 };
 use shared::SessionStatus;
 use std::sync::Arc;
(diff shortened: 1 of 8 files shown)


yumibyte:
Superseded by #1343, opened from the upstream branch `feat/*` on meawoppl/agent-portal instead of the yumibyte fork. Same commits and content.

PR #1393, opened 2026-07-16
Block typing into the composer while in Nav mode

yumibyte:
## Problem

In keyboard Nav mode the composer textarea keeps DOM focus (intentional — so leaving Nav via `Ctrl/Cmd+K` lands the caret ready to type). But it also kept *processing* keystrokes: any key that wasn't a nav shortcut (`e`, `a`, Space, Enter, …) was silently typed into — or submitted from — the composer. The result is a misleading Nav mode: text accumulates in a box you believe is inert.

## Fix

Thread the dashboard's `nav_mode` state down through `SessionView` into `InputBar` and make the composer inert while Nav mode is active:

- **`readonly` textarea** — blocks typing and paste, but (unlike `disabled`) keeps focus and the caret, so leaving Nav mode still lands you ready to type.
- **keydown handler early-returns in Nav mode** with no `prevent_default`/`stop_propagation`, so keys bubble up to `use_keyboard_nav` for navigation, and Enter/history/vim/voice no longer fire from the composer.

Together these guarantee Nav mode is purely navigation: `readonly` blocks character insertion (the default action) and the inert key handler stops the composer's own Enter/history/vim behavior.

### DIFF
--- a/frontend/src/pages/dashboard/page.rs
+++ b/frontend/src/pages/dashboard/page.rs
@@ -717,6 +717,7 @@ pub fn dashboard_page() -> Html {
                                             <SessionView
                                                 session={session.clone()}
                                                 focused={is_focused}
+                                                nav_mode={keyboard_nav.nav_mode}
                                                 on_awaiting_change={on_awaiting_change.clone()}
                                                 on_connected_change={on_connected_change.clone()}
                                                 on_message_sent={on_message_sent.clone()}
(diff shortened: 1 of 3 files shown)


yumibyte:
Superseded by #1391, which fixes the same root cause (composer stays inert while the dashboard is in Nav mode) as part of fixing the number-key pane-switching bug. Closing to avoid a duplicate/conflicting change.

PR #1447, opened 2026-07-20
Render Claude tool progress messages

meawoppl:
## Summary
- route typed Claude `tool_progress` frames through the Claude renderer instead of the raw unknown fallback
- add a compact tool-progress status row showing tool name, elapsed time, parent/task ids, and copyable tool-use id
- re-export the SDK `ToolProgressMessage` type through `shared` and add a regression test for the reported Bash heartbeat payload
- bump workspace version to 2.13.11

## Verification
- `cargo test -p frontend parses_claude_tool_progress_frame_to_claude_renderer`
- `cargo fmt --check`
- `cargo check -p frontend --target wasm32-unknown-unknown`
- `cargo clippy -p frontend --target wasm32-unknown-unknown --all-targets -- -D warnings`
- `python3 ./scripts/check-no-json-macro.py`
- `cd frontend/lint && npm ci && npm run lint:css`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -91,7 +91,7 @@ dependencies = [
 
 [[package]]
 name = "agent-portal"
-version = "2.13.10"
+version = "2.13.11"
 dependencies = [
  "anyhow",
  "chrono",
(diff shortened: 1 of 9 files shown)


meawoppl:
Superseded by #1449. The ephemeral side-channel approach is the better fit for Claude tool_progress heartbeats: it avoids persisting/replaying 30s telemetry frames and gives the UI one live status strip instead of transcript noise. Keeping this closed so the two approaches do not race.

PR #1562, opened 2026-08-06
Add interactive agent sign-in from the Agents matrix

meawoppl:
Signed-out (or unknown-state) installed cells in **Settings ▸ Agents** now
carry a **Sign in** button that drives a launcher-side login flow to
completion, entirely from the browser. Builds on the read-only triage matrix.

## What it does

- **claude**: shows the auth URL, you approve in a browser and paste the code
  back; the launcher submits it and reports the outcome.
- **codex**: shows the device code + verification URL, then polls until the
  browser approval lands.
- On success the matrix re-probes, so the cell flips to "signed in" without a
  manual refresh. Closing the modal mid-flow cancels the launcher-side flow so
  no PTY / app-server lingers.

## Shape

- `shared`: agent-neutral `LoginPresentable` / `LoginInteraction` /
  `AgentLoginOutcome`; four `Start`/`Submit`/`Poll`/`Cancel` launcher protocol
  messages; REST request/response types.
- `claude-session-lib` / `codex-session-lib`: login drivers that park each
  agent's flow (claude's blocking PTY on a thread; codex's device-code watched
  on an async task) and reap it on drop.
- `launcher`: a `flow_id`-keyed `LoginRegistry` and four correlated RPC arms,
  reusing the existing probe request/response correlation.
- `backend`: four **owner-gated** relay endpoints; a dynamic-message 502
  (`AppError::BadGatewayMessage`) so a CLI's own error text reaches the browser.
- `frontend`: `AgentLoginModal` state machine + the matrix buttons.

## Verification

Unit-tested: shared serde round-trips (143), the frontend outcome-status
helper (3), backend error mapping. Full native + WASM builds and clippy are
clean.

⚠️ **The interactive OAuth handshake itself is NOT end-to-end verified** — it
needs a real signed-out agent account + a browser, which this environment
can't drive. Please manually confirm before relying on it:

- [ ] **claude**: a signed-out claude cell → Sign in → URL renders → approve →
      paste code → cell flips to "signed in".
- [ ] **codex**: a signed-out codex cell → Sign in → device code + URL render →
      approve in browser → modal reports success within a couple of polls.
- [ ] Closing the modal mid-flow leaves no orphaned `claude` / `codex app-server`
      process on the launcher host.
- [ ] A start failure (e.g. agent binary missing/broken) surfaces the CLI's own
      error text in the modal, not a generic 500.
- [ ] Only the launcher's owner can drive a login (non-owner → 403/404).

PR #1564, opened 2026-08-06
Add install helpers to the Agents matrix

meawoppl:
A **not-installed** cell in **Settings ▸ Agents** now carries an **Install**
button. It opens a modal showing the exact command that will run on the
launcher host, waits for confirmation, runs it, and re-probes the matrix on
success. Stacked on the sign-in-buttons PR.

## Shape

- `shared`: `AgentType::install_command()` returns structured `program` +
  `args` (a global npm install) — one source of truth used both by the
  launcher (spawned directly, **never** through a shell, so nothing to escape)
  and by the frontend (rendered verbatim in the confirm dialog, so the user
  sees exactly what runs). Plus an `InstallAgent` / `InstallAgentResult`
  launcher protocol pair and an `InstallAgentResponse` REST type.
- `session-lib`: `install::install_agent()` runs the command and relays its
  output tail on failure (npm error, `npm` not on PATH, permissions).
- `launcher`: an `InstallAgent` arm running the install under `spawn_blocking`
  (like the probe).
- `backend`: an **owner-gated** relay endpoint (`POST
  /api/launchers/{id}/agents/{agent}/install`) reusing the probe correlation,
  with a generous 180s timeout for a slow install.
- `frontend`: `AgentInstallModal` (confirm → run → outcome), re-probing on
  success.

## Verification

Unit-tested: shared serde, the `session-lib` output-tail truncation (incl.
multibyte safety), the frontend install-status wording, and that the rendered
command line is the documented installer. Full native + WASM builds, clippy,
and fmt are clean.

⚠️ **The actual install execution is NOT end-to-end verified** here — it needs
a host that's missing an agent CLI and has (or lacks) `npm`. Please manually
confirm:

- [ ] A not-installed cell → Install → the modal shows `npm install -g <pkg>`
      for the right host → Install → success flips the cell to "installed".
- [ ] A failure (no `npm`, or a permissions error) surfaces the command's own
      error text in the modal, not a generic 500.
- [ ] Only the launcher's owner can trigger an install (non-owner → 403/404).

## Note on scope

claude + codex only, matching the current `AgentType`. Muse's install arm
(a piped vendor installer, which slots into the same `program`/`args` shape as
`bash -c "…"`) lands with the Muse integration once `AgentType::Muse` exists.

### DIFF
--- a/backend/src/handlers/launchers.rs
+++ b/backend/src/handlers/launchers.rs
@@ -5,11 +5,12 @@ use axum::{
 use diesel::prelude::*;
 use serde::Deserialize;
 use shared::api::{
-    DirectoryListingResponse, LaunchRequest, ProbeAgentsResponse, StartAgentLoginRequest,
-    StartAgentLoginResponse, SubmitAgentLoginCodeRequest,
+    DirectoryListingResponse, InstallAgentResponse, LaunchRequest, ProbeAgentsResponse,
+    StartAgentLoginRequest, StartAgentLoginResponse, SubmitAgentLoginCodeRequest,
 };
 use shared::{
-    AgentLoginOutcome, LauncherInfo, LauncherToServer, ServerToLauncher, SessionRole, SessionStatus,
+    AgentLoginOutcome, AgentType, LauncherInfo, LauncherToServer, ServerToLauncher, SessionRole,
+    SessionStatus,
 };
 use std::sync::Arc;
 use tracing::{error, info, warn};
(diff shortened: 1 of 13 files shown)


meawoppl:
Kicking CI: this branch has no workflow runs recorded (the PR predates its trigger, or the run was never dispatched), so merging would be blind. Closing/reopening to dispatch `pull_request`.

meawoppl:
**Merging with live verification deferred**, plus one caveat worth recording.

Matt has authorized merge ahead of hands-on testing; he'll exercise install from the Agents matrix against real hosts and file bugs from live use rather than gating the merge.

⚠️ **This branch has no CI runs.** `gh pr checks` reports none, and no workflow runs exist for `meawoppl/agent-install-helpers` — the `pull_request` trigger never dispatched for it (a close/reopen cycle did not kick one either). So unlike #1562 (12/12 green), this merges **without any automated signal**: no clippy, no build, no E2E.

Mitigation: main's own CI runs on push, so a break surfaces immediately post-merge on the main branch rather than silently. Worth a glance at the first main run after this lands, and worth investigating separately why this branch never triggered.

meawoppl:
Superseded by #1566 — identical change rebased onto main (this PR went CONFLICTING once #1562 landed, since both touch the agents-matrix files, and it could not be reopened). #1566 carries the same deferred-live-verification note and, unlike this branch, actually gets CI.

PR #1585, opened 2026-08-06
Lower the frontend render cap to 100 (keep durable retention high)

meawoppl:
The felt UI lag was the **frontend live buffer** — grouping/re-parsing a large buffer on every streamed frame and holding it all as live DOM — not durable history. (Confirmed: prod pins `MESSAGE_RETENTION_COUNT=2000` in env, so the open-path replay was 2000 all along, before and after today's deploy; the speed-up came entirely from the render cap.)

## Change
- `MAX_MESSAGES_PER_SESSION` (frontend render budget): **1000 → 100** — the actual fix.
- `MESSAGE_RETENTION_COUNT` (durable DB history): **stays 1000** — a separate job. Trimming it would discard muse scrollback for zero perf gain. Prod overrides to 2000 via env; this default only governs non-override hosts.

## Why 100 is safe against #1581
Not by budgeting for the flood, but by cutting it at the source: the reminder classifier screen (#1587) `Noop`s muse's `reminder.*` scaffolding (~41–47% of a turn) before the buffer, so a post-screen turn is ~30–40 rendered records.

Supersedes the earlier both-knobs-to-100 approach (#1586) and the 200/three-knob-decouple proposal. A bounded initial-replay cap remains a worthwhile *efficiency* follow-up (stop shipping ~2000 rows to discard most), but it's not the felt lag. Can't merge until Actions recovers.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -912,7 +912,7 @@ When making changes, verify:
 | `PORTAL_STT_VOCABULARY_NAME` | Pre-created custom-vocabulary resource | Optional (`aws`) |
 | `PORTAL_MAX_AUDIO_MB` | Per-recording cap for `POST /api/stt/transcribe` | Optional (default: 25) |
 | `PORTAL_MAX_IMAGE_MB` | Max image size for proxies (also the per-file cap for image `agent-portal show`) | Optional (default: 10) |
-| `PORTAL_MAX_VIDEO_MB` | Per-file cap for videos shown via `agent-portal show` | Optional (default: 100) |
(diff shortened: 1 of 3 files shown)


meawoppl:
Superseded by #1586. After a293998c's retention commit, #1586's head (a825c4b) carries the same render cap 100 + retention 1000 this PR did — verified at the file level. Keeping #1586 as the single canonical cap PR preserves the deployed-image lineage (7571b40) and the #1588 / docker-compose references that point at #1586. Closing this duplicate; no code lost. (Cleanup of the conflicting cap PRs per Matt.)

PR #1603, opened 2026-08-07
Fix platform-native config path handling

### LINKED ISSUE #1591: Config lives under com.anthropic.* bundle ID; installer and launcher disagree on macOS, silently dropping self-hosted backend_url
## Summary

Config lives under a `com.anthropic.*` bundle identifier that has nothing to do with this project, and because `ProjectDirs` expands that identifier differently per platform, the install script and the launcher disagree about where config goes on macOS. The installer's `backend_url` is silently ignored there, and almost every path in `docs/` is Linux-only and wrong on macOS.

Two separable problems — the naming (cosmetic, confusing) and the path mismatch it obscures (an actual bug). Filing together because the fix for one shapes the fix for the other.

## The naming

`launcher/src/config.rs:28`:

```rust
fn config_dir() -> PathBuf {
    directories::ProjectDirs::from("com", "anthropic", "agent-portal")
        .map(|p| p.config_dir().to_path_buf())
        .unwrap_or_else(|| PathBuf::from("/tmp/agent-portal"))
}
```

This is a public repo under `meawoppl/`, unaffiliated with Anthropic, and it writes user auth tokens into a directory named after a company that doesn't ship it. Anyone auditing what's on their machine finds `~/Library/Application Support/com.anthropic.agent-portal/` holding a JWT and reasonably concludes Anthropic put it there. It reads as vendor attribution, not as a namespace.

The literal is duplicated across four call sites, so any rename has to hit all of them:

| File | Line |
|---|---|
| `launcher/src/config.rs` | 29 |
| `proxy/src/config.rs` | 140 |
| `launcher/src/process_manager.rs` | 30 |
| `session-lib/src/output_buffer.rs` | 114 |

## The bug it hides

`directories` 5.0 expands `("com", "anthropic", "agent-portal")` to a genuinely different path per platform, because Linux follows XDG and discards the qualifier and organization entirely:

| Platform | Resulting config dir |
|---|---|
| Linux | `~/.config/agent-portal/` |
| macOS | `~/Library/Application Support/com.anthropic.agent-portal/` |
| Windows | `%APPDATA%\anthropic\agent-portal\...` |

The install script hardcodes the Linux answer for every platform — `backend/src/handlers/downloads.rs:44`:

```bash
CONFIG_DIR="${HOME}/.config/agent-portal"
BIN_NAME="agent-portal"
BIN_PATH="${CONFIG_DIR}/${BIN_NAME}"
CONFIG_FILE="${CONFIG_DIR}/launcher.json"
```

It then writes the server-parameterized `backend_url` into `$CONFIG_FILE`. On macOS the launcher never reads that file. `backend_url` is `Option<String>` (`launcher/src/config.rs:8`) and when absent falls back to `shared::default_backend_url()` — `wss://txcl.io` in release builds.

**So on macOS, `curl https://my-backend/install.sh | bash` installs a launcher pointing at `wss://txcl.io`, not at `my-backend`.** The one piece of configuration the install script exists to deliver is the one that gets dropped. It self-corrects only once the user runs `agent-portal login`, since `save_credentials` writes `backend_url` and `auth_token` together to the real path (`config.rs:107`) — so for the common flow the damage is a confusing first-connect against the wrong host rather than a permanent misconfiguration. For anyone self-hosting, that's still a bad first five minutes with no error message explaining it.

This is exactly what I hit. That machine has a stray `~/.config/agent-portal/launcher.json` containing only `{"backend_url": "wss://portal.cosmicfrontier.org"}`, dated install day, that nothing has ever read — while the live config with the token sits in Application Support. I initially assumed it was a leftover from the `claude-portal` → `agent-portal` migration (CHANGELOG:390); it isn't. The current installer creates it on every macOS install.

Note the binary intentionally lives in `$CONFIG_DIR` too, so `~/.config/agent-portal/` isn't purely vestigial on macOS — it holds the executable that launchd runs. Only the config half is dead. Any fix needs to keep the binary path working or move it deliberately.

## Docs are Linux-only

Nearly every documented path is the Linux one, stated unconditionally:

- `docs/PROXY_AUTH.md:78` — "**Linux/Mac**: `~/.config/agent-portal/config.json`" — explicitly wrong for Mac
- `docs/PROXY_AUTH.md:10`, `:156`, `:163`
- `docs/TESTING.md:131`, `:193`
- `docs/LAUNCHER_ROADMAP.md:23`
- `proxy/src/main.rs:43` — the `--help` text itself

`docs/TROUBLESHOOTING.md:263` is the actively harmful one:

```bash
rm -rf ~/.config/agent-portal/  # Remove cached auth
```

On macOS this deletes the binary and leaves the auth token untouched — the precise opposite of both stated and intended effect. A user following it to fix an auth problem breaks their install and still has the bad token.

`docs/VSCODE_SETUP.md:35` is the only place that gets it right, and it reads like someone hit this and patched one doc.

## Suggested fixes

1. **Rename the qualifier.** `("org", "cosmicfrontier", "agent-portal")` or similar — something that reflects who actually ships this. Needs a migration: on startup, if the new config dir is absent and the old one exists, move it and log the move. Without that, every existing device logs out at once.
2. **Collapse the four literals into one shared helper** (`shared::project_dirs()`), so the identifier exists in exactly one place and this can't drift again.
3. **Make the installer ask the binary where config goes** rather than hardcoding a path — e.g. have it shell out to `agent-portal config-path` (a new subcommand wrapping the existing `config_path_display()` at `config.rs:40`) and write there. That kills the platform mismatch permanently instead of duplicating the platform matrix in bash.
4. **Fix the docs** to give per-platform paths, or point at `agent-portal config-path` and stop enumerating them.

Worth considering whether `ProjectDirs` is the right call at all here. Its whole value is OS-native conventions, but this project wants one predictable location, ships its own binary into the config dir, and has an installer that already assumed `~/.config`. Standardizing on `~/.config/agent-portal/` everywhere (with `%APPDATA%` on Windows) would make the installer, the docs, and the code agree, and would sidestep the naming question entirely since no reverse-DNS identifier would be needed.

meawoppl:
Closes #1591

## Summary
- centralize the native project config directory under the project-owned `org.meawoppl.agent-portal` identifier
- migrate the complete legacy `com.anthropic.agent-portal` directory without overwriting an existing destination
- add `agent-portal config-path` and make the installer query it before writing `backend_url`
- correct macOS config paths and credential cleanup guidance in affected docs

## Verification
- `cargo fmt --all -- --check`
- `cargo test -p session-lib config_paths`
- `cargo test -p agent-portal config`
- `cargo check -p agent-portal -p claude-portal`
- `cargo check -p backend` (with the expected `frontend/dist` directory present)
- `git diff --check`


meawoppl:
Closing in favor of #1601, which fixes the same issue and has now merged. The key difference is migration ordering: the installer creates the destination config/stub before first launch, so #1603's  early return can strand the live legacy credentials. #1601 treats legacy state as authoritative while preserving the installed binary, covering that case.

PR #1640, opened 2026-08-11
Send composer buffer as secure file (#1636)

### LINKED ISSUE #1636: Send composer buffer to the agent as a file (secret-safe path for credentials)
## Problem

There is no way to hand an agent a credential (API key, .env block, token, private key) without it entering the chat log. Anything typed in the composer becomes a `messages` row — persisted, broadcast to connected clients, replayed on reconnect, swept into the long-term archive, and visible in shared read-only sessions. Today the workaround is pasting secrets into chat and hoping retention eventually eats them.

## Proposal

A composer action — e.g. a "send as file" button / `Ctrl+Shift+Enter` variant — that takes the **current message buffer** and delivers it to the agent's host as a **file** instead of a message:

1. Buffer contents ride the existing chunked upload pipeline (#939: web client → backend WS → proxy), never the `AgentInput` path.
2. Proxy writes them to a temp file in the session's environment (e.g. `$XDG_RUNTIME_DIR` or `/tmp/portal-drop-<rand>`, mode `0600`, owner = agent user).
3. The agent receives a small injected notice with **the path only** (e.g. `[file from user: /tmp/portal-drop-a1b2c3, 312 bytes]` plus a one-line reminder not to echo the contents into output).
4. The sender's transcript shows a placeholder card — filename/size/timestamp, no content. Other viewers of a shared session see the same placeholder.

## Non-negotiable secret-handling properties

- Content never touches the `messages` table, WS broadcast to clients, replay, or the session archive.
- Content never appears in backend/proxy logs (the upload path already treats bodies as opaque bytes; verify no debug-format leaks).
- File created `0600`; written atomically (tmp + rename) so no partial world-readable window.
- Composer buffer cleared after send, same as a normal send.
- Size cap (reuse an existing `PORTAL_MAX_*` knob or add `PORTAL_MAX_DROP_KB`, default small — this is for credentials, not datasets).
- Optional but recommended: TTL cleanup of drop files (proxy deletes after N minutes or on session exit), so secrets don't accumulate in /tmp.

## Caveats to document

The mechanism keeps secrets out of the *portal's* durable stores. The agent can still `cat` the file into its own output — the injected notice should tell it not to, and docs should say to prefer `source`/env-var usage over printing.

## Sketch

- `shared/src/endpoints.rs`: reuse `FileUploadStart/Chunk/Result` fields with a `disposition: drop` (or a parallel `BufferDrop` message) so the proxy knows this is a write-to-tmp, not the existing upload flow's destination.
- `frontend` `input_bar.rs`: new action wired to the buffer; placeholder card rendering.
- `proxy`/`session-lib`: tmp-file write + notice injection + TTL cleanup.
- Backend: route-through only; explicitly no persistence.

meawoppl:
Closes #1636

- Reuses typed chunked upload pipeline with `FileUploadDisposition::Drop` (wire-compatible optional `disposition` on `FileUploadStartFields`).
- Backend is route-through only: no `messages` row, WS broadcast/replay, or archive; enforces `PORTAL_MAX_DROP_KB` (default 64 KiB) at start and per-chunk, validates chunk count/bytes, forwards via `SessionManager` with owner-only WS auth; logs never include chunk data.
- Proxy writes atomically (`tmp + rename`) to `$XDG_RUNTIME_DIR` or `/tmp` as `portal-drop-<id>` with `0600` (and `0700` dir), handles duplicate-start cleanup, disconnect/retry idempotency (duplicate indices ignored), injects path-only notice into agent stdin and schedules TTL cleanup (600s), never echoing contents.
- Frontend adds "Send as secure file" action (dropdown + `Ctrl+Shift+Enter`), chunks buffer via `UPLOAD_CHUNK_SIZE`/base64, clears buffer without storing in history/`localStorage`, shows placeholder card (size/timestamp) and error states gated on `FileUploadResult` via `InputBarInbound`.
- Tests: backend upload disposition caps, proxy file handling, frontend placeholder, shared protocol defaults; fmt/clippy clean; no `serde_json::json!` bodies.


meawoppl:
Closing without merge. This stale head conflicts with current main and still has unresolved security/correctness blockers from review: fail-closed capability negotiation in both version-skew directions, client-controlled upload path material, parallel placeholder persistence, false notice-persistence invariant, forbidden dead-code allowances, and process-global environment test races. Secure drop should be reimplemented from current main against the typed PortalContent path rather than reviving this unsafe branch.

Issue #1407, opened 2026-07-17
The Narrator + Illustrator: service sessions producing per-session narration, todos, and SVG storyboards


meawoppl:
A managed, non-displayed agent session pinned to a cheap model (Haiku) that watches the user's other sessions and maintains per session: (1) a single-line "now" summary, (2) a rolling narrative context, (3) the extracted todo list. Feeds the session rail cards, pill menu, `message list`, and the Android status-pill notification + home-screen widget (which today show only name+state).

Depends on #1406 (message peek, incl. the `since` param).

## Architecture — three deliberate splits
1. **Deterministic before generative**: todos come from typed `TodoWrite` (claude) / `update_plan` (codex) tool_use messages — backend parser, zero model cost. Only one-liners + rolling context are LLM work.
2. **A real session, not a backend provider integration**: launcher-managed session with `--model haiku` in claude_args, using the machine's existing CLI auth. The backend continues to hold no provider credentials (rejected alternative: backend calling Anthropic API directly — crosses that line permanently).
3. **Its conversation is the rolling context**: agent context persists across turns; durable outputs are published snapshots.

## Components
1. `sessions.hidden` (server-side) — migration + filtering in session lists, rail, widget/notification, `message list`; still visible in Settings → Sessions. (Today "hidden" is localStorage-only.)
2. Deterministic todo extraction on the message path (same typed seam as push hooks).
3. `session_summaries` table: {session_id PK, one_liner, rolling_context, todos jsonb, updated_at, watermark}.
4. `POST /api/agent/sessions/{id}/summary` (bearer, own-sessions, resolve_user seam).
5. Peek `since` param (in #1406).
6. Narrator runtime: hidden session + scheduled prompt every ~3-5 min: list → peek-since changed sessions → publish → end turn. Skips idle sessions and itself. Kept alive via launcher ExpectedSession reconcile.
7. Consumers (small PRs each): rail card subtitle, pill menu, message list second line, Android widget/notification body, peek header.

## Cost envelope
Haiku + incremental reads + idle-skip: ~10 active sessions x 12 updates/hr x (~2k in / 150 out) ≈ well under $1/day; deterministic todo layer costs nothing and works with the narrator off.

## Phasing
- **P1 (no LLM)**: hidden flag, todo extraction, storage, publish API, rail todos display.
- **P2**: narrator runtime + cron prompt + one-liners/rolling context + web display.
- **P3**: one-liners into Android widget/status pills + peek header.

## Open questions
[...]

PR #1663, opened 2026-08-13
Add the offline illustrator harness

meawoppl:
The deterministic half of the storyboard from #1407 Rev 3 §9: walk a merged PR's commits and emit, per commit, a scene spec plus a rendered SVG frame.

**No model in this version, deliberately.** Rev 3 calls for building the deterministic layer first — file churn, crate grouping, and the commit subject as caption all come from git for free — so that we can judge how much a generative layer actually adds before adding one.

**Scene spec is written alongside the SVG** because it is the durable artifact: the renderer is disposable and history can be re-rendered when it improves. It is also the shape a model would emit once emphasis and captioning go generative, rather than emitting SVG directly.

**Layout is computed once over the union of files the PR touches**, so a file keeps the same rectangle in every frame — that stability is what makes a sequence read as one picture rather than unrelated diagrams.

Runs against already-merged work via `refs/pull/<n>/head`, since the repo squash-merges and deletes branches.

Ran on 12 PRs: 32 frames, 0 invalid SVGs. Findings on frame granularity are in a comment on #1407.

### DIFF
--- a/scripts/illustrator/illustrate.py
+++ b/scripts/illustrator/illustrate.py
@@ -1,0 +1,407 @@
+#!/usr/bin/env python3
+"""Offline illustrator harness for the session storyboard (#1407, Rev 3 P2).
+
+Walks a merged PR's commits and emits, per commit, a *scene spec* (JSON) plus a
+rendered SVG frame. Deliberately split that way: the scene spec is the durable
+artifact and the renderer is disposable, so the whole history can be re-rendered
+when the renderer improves. That is also why the model — when one is eventually
+added to pick emphasis and write captions — will emit a scene spec rather than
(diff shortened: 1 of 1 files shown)


meawoppl:
Closing: wrong approach. This draws where bytes moved — a churn treemap — which cannot explain what a change does. The information needed for an explanatory diagram is not in file/line statistics, so no amount of layout work would have rescued it.

The corollary is that Rev 3 §6 was also wrong: a fixed scene schema with deterministic layout can only express diagrams someone anticipated in advance, and a diagram of a *mechanism* is exactly the case that cannot be anticipated. Replacing both with a prompt for a Fable session that reads the change and composes the SVG directly.

PR #1687, opened 2026-08-14
Add illustrator pipeline tooling

meawoppl:
Tooling around `docs/prompts/illustrate-change.md`, extracted from generating 36 diagrams across two repositories. **Draft** — the prompt and the glyph checker landed separately (#1664, #1666, #1667); this is the surrounding pipeline.

- **`check-layout.py`** — verifies the canonical layout mechanically: canvas, background rect, header and subtitle baselines, `PR #n · title` heading format, the rule, the dashed divider, that content actually fills the panel band, and per-PR `id` prefixing. Sessions run independently and cannot see each other's output, so consistency across a dozen is not something to take on faith.
- **`make-contact-sheet.py`** — tiles diagrams 3x4 with rules between them, **namespacing every `id` per tile**. Bare ids like `arrow-green` collide across files (three of twelve diagrams defined exactly that), and on collision the first definition in document order wins, so later tiles silently draw the wrong marker with nothing erroring.
- **`README.md`** — the generate/check/assemble/export flow, including verifying background opacity by rendering over white, since on a dark viewer transparent and opaque look identical.

Deliberately **not** automated: the diagram itself. An earlier attempt generated them from git statistics and could only ever draw where bytes moved.

One caveat documented in the README: these sessions write a draft, rasterize, look at it, and revise — so a file appearing on disk does not mean the session finished. Assembling on file existence captured a superseded tile twice.

### DIFF
--- a/scripts/illustrator/README.md
+++ b/scripts/illustrator/README.md
@@ -1,0 +1,77 @@
+# Illustrator pipeline
+
+Tooling around [`docs/prompts/illustrate-change.md`](../../docs/prompts/illustrate-change.md),
+which has a Fable session read a PR and compose an SVG explaining what the
+change does. These scripts are the parts worth automating: checking that the
+output follows the canonical layout, and assembling many diagrams into one
+sheet.
+
+The diagram itself is deliberately **not** automated. An earlier attempt
+generated diagrams from git statistics and could only ever draw where bytes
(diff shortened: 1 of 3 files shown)


meawoppl:
Closing the draft as superseded: the effective tooling changes are already present on current main, so this branch has no remaining delta to merge.

Issue #1653, opened 2026-08-13
Muse sessions never get branch or PR metadata updates


meawoppl:
Muse sessions show the branch they were launched with, forever. The PR URL never appears in the pill, and the open-PR list stays empty no matter what the session does.

## Cause

Git metadata refresh — `GitMetadataState`, `GitRefreshTrigger`, and the `get_pr_url` / `get_open_prs` calls that populate them — lives inside `claude-session-lib/src/proxy_session/git_metadata.rs`. Two of the three agent types reach it:

- **claude** natively;
- **codex** because `proxy/src/session.rs:2` re-exports `claude_session_lib::proxy_session::*` wholesale, and `session_event.rs` drives the refresh via `codex_output_has_git_signal`.

**muse** does not. `launcher/src/process_manager.rs:328` constructs `Session::<MuseAgent>` directly, bypassing `proxy_session` entirely. The only git call anywhere on that path is `get_git_branch(&working_directory)` at `process_manager.rs:239`, which runs once at launch.

Note that the underlying helpers in `session-lib/src/git_metadata.rs` are already agent-agnostic — they're pure functions over a working directory. Only the refresh *loop* is behind a claude-named door.

## Fix

Hoist the refresh loop into `session-lib` alongside the pure helpers it already calls, so all three agent types share one implementation. Per-agent output signals (`claude_output_code_change_published`, `codex_output_has_git_signal`) stay in their own crates — after this change they are latency hints that let a session skip the poll interval, not the mechanism itself.

## Acceptance

- Start a muse session, create a branch, commit, and open a PR from it. The pill's branch updates and the PR URL appears **without restarting the session**.
- Verify a `SessionUpdate` carrying the new fields actually reaches the frontend. Compiling is not sufficient evidence here — the whole bug is that the wiring is absent, not that it is wrong.
- claude and codex behavior unchanged.

Prerequisite for the agent-agnostic trigger in #1407.

meawoppl:
First slice of #1657 — the broader relocation of the agent-agnostic proxy session out of `claude-session-lib`. Worth keeping that end state in view while doing this one, but this stands alone and fixes a live muse bug regardless of whether the rest follows.

meawoppl:
Correction after reading the code properly — the stated cause in the description is wrong, and the bug is narrower than filed.

[...]

PR #1697, opened 2026-08-14
Display images an agent reads, again

meawoppl:
Restores the Read-triggered inline image display that #1450 replaced with `agent-portal show`.

`show` is right when an agent *generates* something it wants seen, but it does nothing for the common case of an agent reading a screenshot or a diagram — the user watches it reason about a picture they can't see.

## What's here (backend/proxy — complete)

Built on the current media stack, not the chunked uploader #1451 deleted. Detection lives in a new `media_display` module; the launcher uploads through the same `POST /api/agent/sessions/{id}/media` endpoint `show` uses, so archiving, replay, SVG sanitization (#1530) and eviction are all shared. The detect/side-effect split follows `ClaudeConversationIdSink` — only the launcher holds session credentials, so a standalone proxy passes `None` and doesn't auto-display.

Claude and Codex need separate detectors for the same reason their git signals do (#1653):

- **Claude** keys off the typed `ToolInput::Read`, so an upstream field rename is a compile error rather than a silently-dead detector.
- **Codex** has no typed read tool — it shells out — so its detector parses command argv and is honestly weaker. Gated to `item.completed` (file exists by then) and to a read-ish allowlist, so `rm logo.png` never displays anything.
- **Muse** is excluded: its journal has no command-execution record to key on.

Also extracts `upload_media` from `show`, which resolved the session from its own environment — wrong for the launcher, which manages many sessions at once.

## What's NOT here yet — why this is a draft

The agreed presentation is a **matrix of thumbnails you can click into as a carousel**. That isn't built. Right now each auto-displayed image arrives as its own portal message and renders stacked, one per message.

That matters, because "every read displays" was chosen on the assumption the matrix would contain the noise. Merging this as-is gives the flooding without the mitigation — a directory sweep would stack 30 full-size images down the transcript. **Recommend not merging until the matrix/carousel lands.**

The grouping is the real work: `ImageViewer` already has a single-image lightbox with Escape handling to build the carousel on, but grouping *across* consecutive messages is a transcript-level change, since auto-display produces one message per image.

## Verification

88 tests in claude-session-lib (7 new covering both detectors), 81 in the launcher. Clippy clean. All three binaries build. Not exercised end-to-end against a live session.

### DIFF
--- a/claude-session-lib/src/proxy_session/media_display.rs
+++ b/claude-session-lib/src/proxy_session/media_display.rs
@@ -1,0 +1,190 @@
+//! Auto-display of images an agent reads.
+//!
+//! Reading an image used to render it inline; #1450 removed that in favor of the
+//! explicit `agent-portal show` CLI. `show` is still the right tool when an
+//! agent *generates* something it wants seen, but it does nothing for the common
+//! case of an agent reading a screenshot or a diagram — the user watches it
+//! reason about a picture they cannot see. This restores the implicit path on
+//! top of the current media stack rather than the chunked uploader #1451 deleted:
+//! detection lives here, and the launcher does the upload through the same
+//! `POST /api/agent/sessions/{id}/media` endpoint `show` uses, so archiving,
+//! replay, SVG sanitization (#1530) and eviction behavior are shared.
+//!
+//! Claude and Codex need separate detectors for the same reason their git
+//! signals do (#1653): Claude has a typed `Read` tool, Codex shells out.
+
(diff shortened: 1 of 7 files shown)


meawoppl:
Un-drafted and merging at Matt's call, ahead of the matrix/carousel presentation layer.

Known tradeoff going in: auto-displayed images arrive one per portal message and render stacked, so a directory sweep will put a column of full-size images in the transcript. The matrix grouping is the follow-up that contains that.

